PR #6, opened 2023-01-04
Continuous integration

sylwiaszunejko:
Changes provided by this PR:
* added CI with:
  - run format check
  - install some needed libraries (lld, libunistring-dev, libudev-dev)
  - install seastar
  - build the project
  - run clippy lints
  - run the tests

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -1,0 +1,43 @@
+name: Rust
+
+on:
+  push:
+    branches: [ "main" ]
+  pull_request:
+    branches: [ "main" ]
+
+env:
+  CARGO_TERM_COLOR: always
+
+jobs:
+  format:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+    - name: Format code
+      run: cargo fmt --verbose --all --check
+  build:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+    - name: Install lld
+      run: |
+        sudo apt install lld
+        sudo apt install libunistring-dev
+        sudo apt install libudev-dev
+    - name: Install Seastar
+      run: |
+        cd ..
+        git clone https://github.com/scylladb/seastar.git
+        cd seastar/
+        sudo ./install-dependencies.sh
+        ./configure.py --mode=release --prefix=/usr/local --without-tests --without-demos --without-apps
+        ninja -C build/release
+        sudo ninja -C build/release install
+        export PKG_CONFIG_PATH="/usr/local/seastar/build/release/:$PKG_CONFIG_PATH"
+    - name: Build
+      run: RUSTFLAGS="-C link-arg=-fuse-ld=lld" cargo build
+    - name: Run linter
+      run: cargo clippy --verbose --examples --tests
+    - name: Test
+      run: RUSTFLAGS="-C link-arg=-fuse-ld=lld" cargo test


.github/workflows/rust.yml
```diff
@@ -0,0 +1,42 @@
...
+env:
+  CARGO_TERM_COLOR: always
+
+jobs:
+  format:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+    - name: Format code
+      run: cargo fmt
```
Lorak-mmk: `cargo fmt` formats the files - and formatted files will just be discarded after this job finishes. We need something that either:
- checks the formatting and fails the CI if it's wrong. You can achieve that with `cargo fmt --check`
- formats the files and commits it. I don't like this option (it's more complicated to set up, it creates additional commits making history ugly), so let's stick with `--check`.

.github/workflows/rust.yml
```diff
@@ -0,0 +1,42 @@
+name: Rust
+
+on:
+  push:
+  pull_request:
+    branches: [ "main" ]
```
Lorak-mmk: There are no branches specified in `push` section - is it going to run CI on each push to every branch?
I think we can restrict it to `main` branch, same as with `pull_request` - that's what other Scylla projects do as far as I can see:
https://github.com/scylladb/cpp-rust-driver/blob/9999af005726b9d47d1837536e58d463ec63da63/.github/workflows/build.yml#L4-L7
https://github.com/scylladb/java-driver/blob/3e2d8a1766150d78bd806264ecf1e1870e0f14cf/.github/workflows/tests%40v1.yml#L4-L7
https://github.com/scylladb/scylla-rust-driver/blob/9609a4895a393818e68565d236ebe7460fea6e1c/.github/workflows/cassandra.yml#L5-L8

.github/workflows/rust.yml
Lorak-mmk: scylla-rust-driver and cpp-rust-driver use `cargo fmt --verbose --all -- --check` - let's to the same.
@piodul do you know why there is another double-dash before `--check`? It seems to work the same without it. If there is no difference, let's use `cargo fmt --verbose --all --check`

.github/workflows/rust.yml
```diff
@@ -0,0 +1,42 @@
...
+        cd seastar/
+        sudo ./install-dependencies.sh
+        ./configure.py --mode=release --prefix=/usr/local --without-tests --without-demos --without-apps
+        ninja -C build/release
+        sudo ninja -C build/release install
+        export PKG_CONFIG_PATH="/usr/local/seastar/build/release/:$PKG_CONFIG_PATH"
+    - name: Build
+      run: RUSTFLAGS="-C link-arg=-fuse-ld=lld" cargo build
+    - name: Run linter
+      run: cargo clippy
```
Lorak-mmk: Both scylla-rust-driver and cpp-rust-driver use `--verbose` flag for clippy, let's do the same.
scylla-rust-driver additionally uses `--examples --tests` - I hope we are going to have some examples and tests, so let's use those flags too.
cpp-rust-driver also has `-- -D warnings`: https://github.com/scylladb/cpp-rust-driver/blob/9999af005726b9d47d1837536e58d463ec63da63/.github/workflows/clippy.yml#L19 . I think that's a good idea - that way we'll make sure clippy output is clear, so it will be easier to use locally.

Lorak-mmk (changes requested):

.github/workflows/rust.yml
piodul:
```
piodul@scyllapc ~]$ cargo fmt --help
cargo-fmt
This utility formats all bin and lib files of the current crate using rustfmt.

USAGE:
    cargo fmt [OPTIONS] [-- <rustfmt_options>...]

ARGS:
    <rustfmt_options>...    Options passed to rustfmt
```

It looks like `cargo fmt` calls `rustfmt` underneath. As I understand it, `cargo fmt` might not understand all flags that `rustfmt` accepts so the double-dash can be used to just forward the flags to `rustfmt`.

It indeed looks like `--check` is supported by `cargo fmt`:

```
        --check
            Run rustfmt in check mode
```

so yeah, let's drop the double dash. Maybe `--check` wasn't supported in the past?

.github/workflows/rust.yml
piodul: I think there might be a problem with `-- -D warnings`. Currently, `build.rs` produces a warning - `pkg-config-rs` is unable to parse the `fmt` dependency as it is specified in seastar's pkgconfig file.

If we want to check for warnings in CI (which I agree is a good idea) then we should either fix or slience this warning. At the moment, I don't have a suggestion how to do either of those.

.github/workflows/rust.yml
```diff
@@ -0,0 +1,43 @@
...
+        ./configure.py --mode=release --prefix=/usr/local --without-tests --without-demos --without-apps
+        ninja -C build/release
+        sudo ninja -C build/release install
+        export PKG_CONFIG_PATH="/usr/local/seastar/build/release/:$PKG_CONFIG_PATH"
+    - name: Build
+      run: RUSTFLAGS="-C link-arg=-fuse-ld=lld" cargo build
+    - name: Run linter
+      run: cargo clippy --verbose --examples --tests
+    - name: Test
+      run: RUSTFLAGS="-C link-arg=-fuse-ld=lld" cargo test
```
Lorak-mmk: Please add newline to the end of file

sylwiaszunejko: done

Lorak-mmk (approved):

Lorak-mmk:
One more thing, squash those 2 commits into one

PR #7, opened 2023-01-04
Make API level and scheduling group count configurable

patjed41:
Code added in this PR sets `SEASTAR_API_LEVEL` and `SEASTAR_SCHEDULING_GROUPS_COUNT` the way pkg_config would. Moreover, all other -D variables are also defined. It turned out to be really easy to implement. It's just a simple iteration over the elements of `seastar.defines`.

I also added `.cpp_link_stdlib("stdc++")` that will be necessary later.

### DIFF
--- a/seastar/build.rs
+++ b/seastar/build.rs
@@ -30,14 +30,17 @@ fn main() {
         .map(|p| PathBuf::try_from(p).unwrap())
         .collect::<Vec<_>>();
 
-    // TODO: The API level and scheduling group count
-    // should be configurable somehow
-    cxx_build::bridges(&cxx_bridges)
+    let mut build = cxx_build::bridges(&cxx_bridges);
+    for (var, value) in &seastar.defines {
+        match value {
+            Some(val) => build.define(var, val.as_str()),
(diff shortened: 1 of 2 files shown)


seastar/build.rs
```diff
@@ -30,15 +30,19 @@ fn main() {
         .map(|p| PathBuf::try_from(p).unwrap())
         .collect::<Vec<_>>();
 
-    // TODO: The API level and scheduling group count
-    // should be configurable somehow
-    cxx_build::bridges(&cxx_bridges)
+    let mut build = cxx_build::bridges(&cxx_bridges);
+    for (var, value) in &seastar.defines {
+        match value {
+            Some(val) => build.define(var, &val[..]),
```
Lorak-mmk: `&val[..]` -> `val.as_str()`

Lorak-mmk (changes requested):
Left one comment, otherwise it's good to go.

seastar/build.rs
patjed41: Done.

Lorak-mmk (approved):

Issue #2, opened 2022-12-22
Create a macro like tokio::test


piodul:
(Depends on zpp-2022-rust-seastar/seastar-rs#1)

We would like to make it easy to write tests in Rust that run on Seastar runtime. For example, the Tokio runtime has such a macro: https://docs.rs/tokio/latest/tokio/attr.test.html

I imagine that writing tests in Rust should be as easy as:

```rust
#[seastar::test]
async fn test_frobnication() {
    // You can use seastar functions here without any problems, and await on them
    do_frobnicate().await;
}
```

Let's start with a very simple procedural macro (like described above) which does not take any arguments.

Issue #1, opened 2022-12-22
Make it possible to configure and start seastar from Rust


piodul:
We would like to allow Rust to start the Seastar runtime.

In C++, the process of starting the runtime has a small number of steps:

- First, `app_template::seastar_options` is created and set appropriately
- The `app_template` is created, with seastar options passed as the first argument
- Finally, `app_template::run` is called to start the application

We would like to offer a similar experience in Rust:

- `Options`: for now let's only make it possible to set the application name, description and the number of shards. The properties should be settable via setters/getters.
- `AppTemplate`: it should be possible to create it from `Options` and it should have only one method `run` which works analogously as the original `app_template::run`.

kfernandez31:
I can't edit assignees, so I hope this comment suffices - I'm assigning myself.

piodul:
Fixed in https://github.com/zpp-2022-rust-seastar/seastar-rs/pull/23 .

Issue #5, opened 2023-01-04
Add a way to spawn tasks on the local shard in Rust, similarly to tokio::spawn


cvybhu:
[`tokio::spawn`](https://docs.rs/tokio/latest/tokio/fn.spawn.html)  is a method that allows to spawn a new asynchronous task in the background.

The task is completely independent - subsequent operations done by the caller are done in parallel and aren't synchronized with the task in any way.

It should be possible to asynchronously wait for the task to finish.
The task might return a result which should also be accessible to the caller.

```rust
async fn main() {
    let task: JoinHandle<i32> = tokio::spawn(long_computation());

    // Compute something in parallel to long_computation()
    let computed: i32 = other_long_computation().await;

    // Wait for the task to finish
    let task_res: i32 = task.await.unwrap();

    let result: i32 = computed + task_res;
}

```

The big difference with `tokio` is that `seastar` has the shard-per-core architecture and each task is executed on a single shard. That means that the future doesn't have to be `Send` or `Sync`. In tokio `Send` and `Sync` are a must because the tasks can be migrated between cores.

Some useful seastar functions:
* `make_task`
* `schedule` - usually isn't used manually but maybe this time it can

PR #8, opened 2023-01-04
Gate

### LINKED ISSUE #4: Expose seastar::gate to Rust
https://github.com/scylladb/seastar/blob/master/include/seastar/core/gate.hh

The seastar::gate is a very simple synchronization primitive. Its purpose is to enable graceful shutdown for resources that are accessed concurrently.

A gate starts in an "open" state. While it is open, it can be "entered" and "left" by any number of concurrent operations.
At some point, a gate can be closed. Closing the gate prevents more operations from entering the gate and waits until the operations that are in the gate leave. After the gate is closed, one can be sure that there is nobody "inside" it.

We would like to be able to provide an interface that allows using seastar::gates from C++. We would like to support the following operations:

```rust
impl Gate {
    // Creates a new gate.
    // The gate has a non-trivial move constructor, so we can't return it directly by value.
    // This function will probably be most useful in tests.
    pub fn new() -> Pin<Box<Gate>>;

    // Tries to enter the gate.
    // If it succeeds, it returns an object that will leave the gate when destroyed (RAII).
    // If it fails, it returns an Err variant (a cpp exception, probably)
    pub fn try_enter(&'a self) -> Result<GateHolder<'a>>;

    // Closes the gate.
    // It must be called at most once (the underlying implementation aborts if it is closed twice).
    // Returns a future that resolves after everybody have left the gate.
    async fn close(&self);
}
```

patjed41:
Fixes: #4
Depends on: #1, #2, #13

This PR introduces exposition of `seastar::gate` to Rust as `struct Gate` with methods:
```rust
pub fn new() -> Pin<Box<Self>>;
pub fn try_enter(&'a self) -> Result<GateHolder<'a>, GateClosedError>;
pub async fn close(&self);
```
There are also 2 auxiliary types:
- `GateHolder<'a>` – equivalent of `seastar::gate::holder`,
- `GateClosedError` – equivalent of `seastar::gate_closed_exception`.

Currently, there are 3 tests testing the following scenarios:
- `close`,
- `try_enter -> leave -> close`,
- `close -> try_enter` (`GateClosedError` is returned).

A very important case `try_enter -> close -> leave` (`close` waits) hasn't been tested yet, because it requires Seastar runtime. This particular case (and maybe some other cases with more enters) will be tested when Seastar runtime is exposed to Rust.

There are some issues with this PR that I have noticed:
- `GateHolder` holds reference to `gate` just to implement the lifetime. It would be really nice if something like this
  ```rust
  pub struct GateHolder<'a> {
      _holder: 'a UniquePtr<gate_holder>,
  }
  ```
  worked, but unfortunately it doesn't. I haven't found a way to implement `GateHolder` without keeping `&'a gate`.
- Our fork of `cxx-async` is not usable yet, because PR hasn't been accepted  there. Instead, I used fork of our fork.
- `cxx-async` throws many warnings.
- `CloseGateFuture` appears in the documentation generated by `cargo doc` and I don't know how to hide it. This
  ```rust
  #[doc(hidden)]
  type CloseGateFuture = crate::CloseGateFuture;
  ```
  doesn't work.
- `GateClosedError` has a hard coded implemantation of the `Debug` trait. It should be possible to print `seastar::gate_closed_exception::what()` somehow. I'll do it if it is necessarry.

### DIFF
--- a/seastar/src/gate.rs
+++ b/seastar/src/gate.rs
@@ -1,0 +1,8 @@
+#[cxx::bridge(namespace = "seastar_ffi::gate")]
+mod ffi {
+    unsafe extern "C++" {
+        include!("seastar/src/gate.hh");
+    }
+}
+
+use ffi::*;
--- a/seastar/src/gate.rs
+++ b/seastar/src/gate.rs
@@ -1,8 +1,36 @@
+use cxx::UniquePtr;
+
 #[cxx::bridge(namespace = "seastar_ffi::gate")]
 mod ffi {
     unsafe extern "C++" {
         include!("seastar/src/gate.hh");
+
+        type gate;
+
+        fn new_gate() -> UniquePtr<gate>;
     }
 }
 
 use ffi::*;
+
+/// Facility to stop new requests, and to tell when existing requests are done.
+///
+/// When stopping a service that serves asynchronous requests, we are faced with
+/// two problems: preventing new requests from coming in, and knowing when existing
+/// requests have completed. The `Gate` class provides a solution.
+pub struct Gate {
+    inner: UniquePtr<gate>,
+}
+
+impl Default for Gate {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+impl Gate {
+    /// Creates a new gate.
+    pub fn new() -> Self {
+        Gate { inner: new_gate() }
+    }
+}
--- a/seastar/src/gate.rs
+++ b/seastar/src/gate.rs
@@ -88,3 +88,97 @@ impl<'a> GateHolder<'a> {
         }
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate as seastar;
+    use futures::join;
+    use std::{cell::RefCell, rc::Rc};
+
+    #[seastar::test]
+    async fn test_gate_only_close() {
+        let gate = Gate::new();
+        gate.close().await;
+    }
+
+    #[seastar::test]
+    async fn test_gate_leave_then_close() {
+        let gate = Gate::new();
+
+        {
+            let _ = gate.try_enter().unwrap();
+        }
+
+        gate.close().await;
+    }
+
+    #[seastar::test]
+    async fn test_gate_close_then_leave() {
+        let gate = Gate::new();
+
+        let closing_started = Rc::new(RefCell::new(false));
+        let closing_finished = Rc::new(RefCell::new(false));
+
+        let handler = gate.try_enter().unwrap();
+        let leave_future = async {
+            drop(handler);
+            assert!(*closing_started.borrow() && !*closing_finished.borrow());
+        };
+
+        let close_future = async {
+            *closing_started.borrow_mut() = true;
+            gate.close().await;
+            *closing_finished.borrow_mut() = true;
+        };
+
+        // join! tries to finish close_future first.
+        join!(close_future, leave_future);
+        assert!(*closing_finished.borrow());
+    }
+
+    #[seastar::test]
+    async fn test_gate_many_leave() {
+        let gate = Gate::new();
+
+        let closing_finished = Rc::new(RefCell::new(false));
+
+        let handler1 = gate.try_enter().unwrap();
+        let leave_future1 = async {
+            drop(handler1);
+            assert!(!*closing_finished.borrow());
+        };
+
+        let handler2 = gate.try_enter().unwrap();
+        let leave_future2 = async {
+            drop(handler2);
+            assert!(!*closing_finished.borrow());
+        };
+
+        let handler3 = gate.try_enter().unwrap();
+        let leave_future3 = async {
+            drop(handler3);
+            assert!(!*closing_finished.borrow());
+        };
+
+        let close_future = async {
+            gate.close().await;
+            *closing_finished.borrow_mut() = true;
+        };
+
+        join!(leave_future1, close_future, leave_future2, leave_future3);
+        assert!(*closing_finished.borrow());
+    }
+
+    #[seastar::test]
+    async fn test_gate_close_then_enter() {
+        let gate = Gate::new();
+
+        gate.close().await;
+
+        match gate.try_enter() {
+            Err(GateClosedError) => (),
+            _ => panic!("gate.try_enter() should return Err(GateClosedError)."),
+        }
+    }
+}
--- a/seastar/src/gate.rs
+++ b/seastar/src/gate.rs
@@ -1,4 +1,6 @@
 use cxx::UniquePtr;
+use std::marker::PhantomData;
+use thiserror::Error;
 
 #[cxx::bridge(namespace = "seastar_ffi::gate")]
 mod ffi {
@@ -6,12 +8,19 @@ mod ffi {
         include!("seastar/src/gate.hh");
 
         type gate;
+        type gate_holder;
 
         fn new_gate() -> UniquePtr<gate>;
+        fn new_gate_holder(gate: &UniquePtr<gate>) -> Result<UniquePtr<gate_holder>>;
     }
 }
 
 use ffi::*;
+
+/// Error returned by [`try_enter`](Gate::try_enter) when called on closed gate.
+#[derive(Error, Debug)]
+#[error("GateClosedError: gate closed")]
+pub struct GateClosedError;
 
 /// Facility to stop new requests, and to tell when existing requests are done.
 ///
@@ -32,5 +41,36 @@ impl Gate {
     /// Creates a new gate.
     pub fn new() -> Self {
         Gate { inner: new_gate() }
+    }
+
+    /// Tries to enter the gate.
+    ///
+    /// If it succeeds, it returns [`GateHolder`] that will leave the gate when destroyed (RAII).
+    ///
+    /// If it fails, it returns [`GateClosedError`].
+    pub fn try_enter(&self) -> Result<GateHolder, GateClosedError> {
+        GateHolder::new(&self.inner)
+    }
+}
+
+/// Facility to hold a gate opened using RAII.
+///
+/// A [`GateHolder`] is obtained when [`try_enter`](Gate::try_enter) succeeds.
+///
+/// The [`Gate`] is left when the [`GateHolder`] is dropped.
+pub struct GateHolder<'a> {
+    _inner: UniquePtr<gate_holder>,
+    _phantom: PhantomData<&'a ()>,
+}
+
+impl<'a> GateHolder<'a> {
+    fn new(gate: &'a UniquePtr<gate>) -> Result<Self, GateClosedError> {
+        match new_gate_holder(gate) {
+            Ok(holder) => Ok(GateHolder {
+                _inner: holder,
+                _phantom: PhantomData,
+            }),
+            Err(_) => Err(GateClosedError),
+        }
     }
 }
--- a/seastar/src/gate.rs
+++ b/seastar/src/gate.rs
@@ -10,8 +10,12 @@ mod ffi {
         type gate;
         type gate_holder;
 
+        #[namespace = "seastar_ffi"]
+        type VoidFuture = crate::cxx_async_futures::VoidFuture;
+
         fn new_gate() -> UniquePtr<gate>;
         fn new_gate_holder(gate: &UniquePtr<gate>) -> Result<UniquePtr<gate_holder>>;
+        fn close_gate(gate: &UniquePtr<gate>) -> VoidFuture;
     }
 }
 
@@ -50,6 +54,16 @@ impl Gate {
     /// If it fails, it returns [`GateClosedError`].
     pub fn try_enter(&self) -> Result<GateHolder, GateClosedError> {
         GateHolder::new(&self.inner)
+    }
+
+    /// Closes the gate.
+    ///
+    /// It must be called at most once (the underlying implementation aborts if it is closed twice).
+    ///
+    /// Returns a future that resolves after everybody have left the gate.
+    pub async fn close(&self) {
+        crate::assert_runtime_is_running();
+        close_gate(&self.inner).await.unwrap();
     }
 }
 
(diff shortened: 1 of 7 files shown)


patjed41:
I was worried about `CloseGateFuture` appearing in the documentation but the problem here is that it's public. Change like this
```rust
type CloseGateFuture = super::CloseGateFuture;
```
plus changing
```rust
pub use gate::*;
```
to
```rust
pub use gate::{GateClosedError, Gate, GateHolder};
```
in `lib.rs` solves it. But still, it feels bad to specify all types publicly imported from a module in `lib.rs` just to avoid exposition of a single type used by cxx-async. I would appreciate some suggestions about this issue.

seastar/src/gate.rs
```diff
@@ -0,0 +1,136 @@
...
+}
+
+impl<'a> Gate {
+    /// Creates a new gate.
+    ///
+    /// The underlying `gate` type has a non-trivial move constructor,
+    /// so we can't return it directly by value.
+    pub fn new() -> Pin<Box<Self>> {
+        Box::pin(Gate { gate: new_gate() })
+    }
```
Lorak-mmk: `struct Gate` holds UniquePtr to `gate` instead of holding it directly, so putting another `Box` creates a heap indirection - is it really necessary? It's possible that I don't understand something, but why not:
1. Return `Self` from this function - it looks like `UniquePtr` can be freely moved around, so no need to pin it, and it already pins the value inside, so we can't move it out: `https://docs.rs/cxx/latest/cxx/struct.UniquePtr.html#method.as_mut`. That would be my preferred approach.
2. Hold the gate directly in the struct, without UniquePtr - then we actually need to use `Pin<Box<Self>>`

seastar/src/gate.rs
```diff
@@ -0,0 +1,136 @@
...
+
+    /// Closes the gate.
+    ///
+    /// It must be called at most once (the underlying implementation aborts if it is closed twice).
+    ///
+    /// Returns a future that resolves after everybody have left the gate.
+    pub async fn close(&self) {
+        close_gate(&self.gate).await.unwrap();
+    }
```
Lorak-mmk: After reading this now, I'm wondering: is it possible to create an implementation that leverages Rust's type system to prevent calling close twice?
We could return 2 objects from `Gate::new()` - one would be used to enter the gate, and would be shared, another would be used to close the gate (`GateCloser`?) and would not be shared. That way the close method could take it by value, making it impossible to call it twice.

Lorak-mmk (changes requested):

seastar/src/gate.rs
patjed41: I'm not sure if approach 2 can work. The gate is of type `gate` which is an opaque type from C++ so it cannot be moved by value across the ffi. Therefore, we are forced to create it on the Rust side. I don't think it's possible (how are we going to call a C++ constructor?).

Approach 1 should be correct. I considered description of this issue as something that shouldn't be changed, but after putting `gate` in UniquePtr it makes sense to change return type to `Self`. I'll fix this in a moment.

Lorak-mmk: Imo issue description are absolutely subject to change - the aim of the project is to create an ergonomic API, so if you can improve the API then by all means do it.

seastar/src/gate.rs
patjed41: It's definitely possible. I've just tried it and manged to do it by holding the gate in SharedPtr (instead of UniquePtr) by `Gate` and `GateCloser`. This is how it works:
```rust
pub struct Gate {
    gate: SharedPtr<gate>,
}

pub struct GateCloser {
    gate: SharedPtr<gate>,
}

impl<'a> Gate {
    pub fn new() -> (Self, GateCloser) {
        let new_gate = Gate { gate: new_gate() };
        let gate_closer = GateCloser { gate: new_gate.gate.clone() };
        (new_gate, gate_closer)
    }
   
    pub fn try_enter(&'a self) -> Result<GateHolder<'a>, GateClosedError> {
        GateHolder::new(&self.gate)
    }
}

impl GateCloser {
    pub async fn close(self) {
        close_gate(&self.gate).await.unwrap();
    }
}
```
I tried some other implementations:
1. `Gate` still holds `UniquePtr<gate>` and `GateCloser` holds `& UniquePtr<gate>`.
2. `Gate` still holds `UniquePtr<gate>` and `GateCloser` holds `& gate`.

I couldn't make the first one work because I couldn't implement `Gate::new()`. There were compilation errors like "returns a value referencing data owned by the current function". The second one is probably hopeless because calling `seastar::gate::close()` across ffi requires `&mut gate` instead of `& gate`. So the only solution I can think of is `SharedPtr`.

I'm also not sure if preventing multiple calls of `close` like this is really worth it. It complicates our simple interface quite a bit.

piodul: We discussed this recently with @Lorak-mmk and we don't think we should follow this approach. While the API is slightly nicer, the gate still can be shared with C++ code so this doesn't prevent double close.

Perhaps we could implement a pure-Rust `Gate`; it would make sense for pure Rust applications (i.e. not a component of an existing project). However, in this issue let's concentrate on defining an API for cooperation with a C++ type.

Note: if the gate is closed twice then the second close will abort the process - but it's _still a safe behavior_ according to Rust's definition of safety.

piodul:
> This PR introduces exposition of `seastar::gate` to Rust as `struct Gate` with methods:
> 
> ```rust
> pub fn new() -> Pin<Box<Self>>;
> pub fn try_enter(&'a self) -> Result<GateHolder<'a>, GateClosedError>;
> pub async fn close(&self);
> ```
> 
> There are also 2 auxiliary types:
> 
> * `GateHolder<'a>` – equivalent of `seastar::gate::holder`,
> * `GateClosedError` – equivalent of `seastar::gate_closed_exception`.
> 
> Currently, there are 3 tests testing the following scenarios:
> 
> * `close`,
> * `try_enter -> leave -> close`,
> * `close -> try_enter` (`GateClosedError` is returned).
> 
> A very important case `try_enter -> close -> leave` (`close` waits) hasn't been tested yet, because it requires Seastar runtime. This particular case (and maybe some other cases with more enters) will be tested when Seastar runtime is exposed to Rust.
> 
> There are some issues with this PR that I have noticed:
> 
> * `GateHolder` holds reference to `gate` just to implement the lifetime. It would be really nice if something like this
>   ```rust
>   pub struct GateHolder<'a> {
>       _holder: 'a UniquePtr<gate_holder>,
>   }
>   ```
>
>   worked, but unfortunately it doesn't. I haven't found a way to implement `GateHolder` without keeping `&'a gate`.

You can use [`PhantomData`](https://doc.rust-lang.org/std/marker/struct.PhantomData.html). It is very helpful when you want to implement a generic type that has a type or lifetime parameter but is actually not necessary in the struct definition.

Here, I think it should look something like this:

```rust
use std::marker::PhantomData;
pub struct GateHolder<'a> {
    _holder: UniquePtr<gate_holder>,
    _phantom: PhantomData<&'a ()>,
}
```

> * Our fork of `cxx-async` is not usable yet, because PR hasn't been accepted  there. Instead, I used fork of our fork.
> * `cxx-async` throws many warnings.
> * `CloseGateFuture` appears in the documentation generated by `cargo doc` and I don't know how to hide it. This
>   ```rust
>   #[doc(hidden)]
>   type CloseGateFuture = crate::CloseGateFuture;
>   ```
>     
>   doesn't work.

I suppose that the `#[doc(hidden)]` parameter should appear next to where the original `CloseGateFuture` is defined. I'm not sure whether this is possible to do, it probably depends on how the `cxx_async` macro was implemented. Maybe you can put it before `unsafe impl Future for CloseGateFuture`?

> * `GateClosedError` has a hard coded implemantation of the `Debug` trait. It should be possible to print `seastar::gate_closed_exception::what()` somehow. I'll do it if it is necessarry.

I suggest adding a dependency on the [`thiserror`](https://docs.rs/thiserror/latest/thiserror/) crate. It makes generating such error types very easy.

----

Last thing: if you send a PR that should close an issue, please add the following line to the PR description:

```
Fixes: #issue_number
```

piodul:
> I was worried about `CloseGateFuture` appearing in the documentation but the problem here is that it's public. Change like this
> 
> ```rust
> type CloseGateFuture = super::CloseGateFuture;
> ```
> 
> plus changing
> 
> ```rust
> pub use gate::*;
> ```
> 
> to
> 
> ```rust
> pub use gate::{GateClosedError, Gate, GateHolder};
> ```
> 
> in `lib.rs` solves it. But still, it feels bad to specify all types publicly imported from a module in `lib.rs` just to avoid exposition of a single type used by cxx-async. I would appreciate some suggestions about this issue.

I suggest introducing an internal module just for `cxx_async`'s futures. Their names would correspond to the types that they return, so it won't feel as awkward to define it in a different place than they are used. Moreover, such a single type could be reused in multiple places.

patjed41:
>You can use [PhantomData](https://doc.rust-lang.org/std/marker/struct.PhantomData.html). It is very helpful when you want to implement a generic type that has a type or lifetime parameter but is actually not necessary in the struct definition.

It worked nicely.

>I suggest adding a dependency on the [thiserror](https://docs.rs/thiserror/latest/thiserror/) crate. It makes generating such error types very easy.

I used it. Putting `#[error("GateClosedError: gate closed")]` before the `GateClosedError` declaration sets the `Display` message instead of the `Debug` one. I guess it's fine?

>I suggest introducing an internal module just for cxx_async's futures. Their names would correspond to the types that they return, so it won't feel as awkward to define it in a different place than they are used. Moreover, such a single type could be reused in multiple places.

Done. That was a very good idea. It solved all the issues at the same time.

patjed41:
I've made some changes:
- C++ side of the ffi uses `seastar_ffi::gate` namespace,
- tests use `seastar::test` macro,
- there is a new test that checks the case `try_enter -> close -> leave` (there was a segfault without the runtime),
- `Gate::close` checks if the runtime is running using `assert_runtime_is_running!` macro.

I think that this task is finished now and it is only waiting for a review. If tests are not strong enough, I can think of writing more. The problem is that I don't know how to do it reasonably. Spawning tasks that would leave the gate after some time could be a solution, but #5 hasn't been finished yet.

seastar/src/gate.rs
```diff
@@ -28,9 +37,40 @@ impl Default for Gate {
...
     }
+
+    /// Tries to enter the gate.
+    ///
+    /// If it succeeds, it returns [GateHolder] that will leave the gate when destroyed (RAII).
+    ///
+    /// If it fails, it returns [GateClosedError].
+    pub fn try_enter(&'a self) -> Result<GateHolder<'a>, GateClosedError> {
+        GateHolder::new(&self.gate)
+    }
```
Lorak-mmk: Can't the `try_enter` method be liftime-generic instead of whole impl block?

seastar/src/gate.rs
```diff
@@ -56,10 +56,12 @@ impl<'a> Gate {
...
-    // Closes the gate.
-    // It must be called at most once (the underlying implementation aborts if it is closed twice).
-    // Returns a future that resolves after everybody have left the gate.
-    async fn close(&self) {
+    /// Closes the gate.
+    ///
+    /// It must be called at most once (the underlying implementation aborts if it is closed twice).
+    ///
+    /// Returns a future that resolves after everybody have left the gate.
+    pub async fn close(&self) {
```
Lorak-mmk: This change looks like it's in the wrong commit - it should be squashed into the commit that added close function.

Lorak-mmk (changes requested):

seastar/src/gate.rs
patjed41: Of course it can. Fixed.

seastar/src/gate.rs
patjed41: Yes, that was an unintended oversight.

patjed41:
Update:
- Made changes requested in a review.
- Added more verbose commit messages to commits where it seemed reasonable.
- Split *Add Gate::new* commit to 2 commits: *Create gate module* and *Add Gate with its constructor*. I thought that this commit did too many things.
- Moved cherry-picking API safety to a separate commit.
- Changed some names to better match Rust conventions:
  - `Gate::gate` -> `Gate::inner`,
  - `GateHolder::_holder` -> `GateHolder::_inner`.

seastar/src/gate.rs
```diff
@@ -88,3 +88,80 @@ impl<'a> GateHolder<'a> {
...
+        let leave_future3 = async move {
+            drop(handler3);
+        };
+
+        let close_future = async {
+            gate.close().await;
+        };
+
+        join!(leave_future1, close_future, leave_future2, leave_future3);
+    }
```
piodul: Those test cases should verify that the `leave_future{1,2,3}` have completed and dropped their handlers before `gate.close()` resolves.

piodul (commented):
Looks good from my side, apart from a single comment. Let's schedule this for merge after https://github.com/zpp-2022-rust-seastar/seastar-rs/pull/22 - a slight amount of conflict resolution will be needed.

@Lorak-mmk please review, too - you left a "changes requested" review so I can't merge until you approve.

seastar/src/gate.rs
patjed41: I did that. I also added checking in `test_gate_close_then_leave` that `close_future` is really polled first by `join!`.

patjed41:
Small update:
- I noticed that lifetimes in `Gate::try_enter` can be elided and I did that.
- I fixed an *unused import* warning in `submit_to.rs` (last commit). I think this PR will be merged first so it looks like a good place for a fix like that.

seastar/src/gate.rs
```diff
@@ -0,0 +1,184 @@
...
+
+        gate.close().await;
+    }
+
+    #[seastar::test]
+    async fn test_gate_close_then_leave() {
+        let gate = Gate::new();
+
+        let closing_started = Rc::new(RefCell::new(false));
+        let closing_finished = Rc::new(RefCell::new(false));
```
piodul, line 121: Tip for the future: for very simple `Copy` types which you don't need to borrow but rather copy/overwrite, it's slightly better to use `Cell`: https://doc.rust-lang.org/std/cell/struct.Cell.html .

piodul (approved):
LGTM

Lorak-mmk (approved):

PR #10, opened 2023-01-14
Configuring and starting Seastar

kfernandez31:
This PR introduces the `AppTemplate` and `Options` structs as described in #1. To call Seastar's `app_template::run` under the hood, one must do so through `AppTemplate::run_{void, int}`, which accept function pointers. To make them accept an `async` block, one can do:

```rust
let opts = Options::default;
// set some options here
let app = AppTemplate::new_from_options(opts);
app.run_void(
    futures::executor::block_on(async { 
            some_async_work().await 
        }
);
```

## Notes:
- each `seastar-rs`-based app must have the above code, and so it'd be convenient to turn it into a macro. Luckily, we have an issue for that.
- each call of `AppTemplate::run_{void, int}` spawns a new thread. We might want to remove that slight overhead somehow.


piodul:
> ```rust
> app.run_void(
>     futures::executor::block_on(async { 
>             some_async_work().await 
>         }
> );
> ```

This won't work.

`futures::executor::block_on` is a function that does not return until the rust future provided to it becomes ready. It basically steals the whole thread just for this particular rust future. This is bad because the runtime thread maintains a queue of tasks to execute and has to do some maintenance regularly, i.e. check timers and poll for I/O - so this prevents other tasks from being run and I/O from working correctly. In seastar, each task is supposed to execute for a small amount of time and then yield control to the runtime so that it gives a chance for other things to be done.

The original `app_template::run` takes a function which returns a `seastar::future<>`. You should use `cxx_async` in order to "convert" the rust future to a seastar future.

I guess the call to `app_template::run` should look a bit like this:

```c++
// `main_rust_future` is of future type generated by cxx_async
exit_value = app->run(ac, av, [&] () -> seastar::future<> {
    co_await main_rust_future;
});
```

On the other hand, the `run_void function should just take a future that does not return anything when resolved:

```rust
pub fn run_void(&self, args: &Vec<String>, fut: impl Future<Output=cxx_async::CxxAsyncResult<()>> + 'static) -> i32 {
    // `VoidFuture` type was generated by cxx_async
    // Perhaps you will have to work around the fact here that VoidFuture requires the future to be Send
    run_void(&self.app, args, VoidFuture::fallible(fut))
}
```

seastar/build.rs
```diff
@@ -43,6 +44,7 @@ fn main() {
         .flag_if_supported("-fcoroutines")
         .includes(&seastar.include_paths)
         .cpp_link_stdlib("stdc++")
+        .file("src/config_and_start_seastar.cc")
```
piodul: How about a global static such as CXX_BRIDGES that will specify all the cpp files that the project needs to compile?

seastar/src/config_and_start_seastar.cc
```diff
@@ -0,0 +1,110 @@
+#include "config_and_start_seastar.hh"
+
+namespace seastar {
```
piodul: Please do not put things directly into the `seastar` namespace. You are risking name conflicts with items that might be introduced later by seastar. For example, `seastar_ffi` might be more appropriate.

Besides, lots of functions have names that are very likely to collide with items that we might try to introduce later (for example: `set_name`, `set_smp`). As `seastar_ffi` will be a root namespace of the project, I'm not sure if it is a good idea to put them there directly. I would introduce another namespace for those items (e.g. `seastar_ffi::configure_and_start_seastar`).

seastar/src/config_and_start_seastar.cc
```diff
@@ -0,0 +1,110 @@
...
+        const rust::String& arg = args[i];
+        av[i] = (char*) calloc(arg.size() + 1, sizeof(char));
+        if (av[i] == nullptr) {
+            return nullptr;
+        }
+        strncpy(av[i], arg.data(), arg.size());
+    }
+
+    return av;
+}
```
piodul: Why are you using calloc? We are using C++. Why not use `new char*[args.size() + 1]` etc.?

IMO we should use Rust to convert the string arguments to the right representation and then pass it to C++. It should be shorter and less prone to memory leaks.

seastar/src/config_and_start_seastar.hh
```diff
@@ -0,0 +1,30 @@
...
+
+rust::Str get_description(const std::unique_ptr<seastar_options>& opts);
+
+uint32_t get_smp(const std::unique_ptr<seastar_options>& opts);
+
+void set_name(const std::unique_ptr<seastar_options>& opts, const rust::Str name);
+
+void set_description(const std::unique_ptr<seastar_options>& opts, const rust::Str description);
+
+void set_smp(const std::unique_ptr<seastar_options>& opts, const uint32_t smp);
```
piodul: Why not pass `seastar_options&`?

seastar/src/config_and_start_seastar.rs
```diff
@@ -0,0 +1,293 @@
...
+    /// use seastar::{AppTemplate, Options};
+    ///
+    /// let func = || { println!("{}", 42); };
+    ///
+    /// let app = AppTemplate::default();
+    /// let args = vec!["hello".to_owned()];
+    ///
+    /// assert_eq!(app.run_void(&args, func), 0);
+    /// ```
+    pub fn run_void(&self, args: &Vec<String>, func: fn()) -> i32 {
```
piodul: Instead of passing a reference to a vector, it's slightly better to pass a slice. It allows for more use cases (you can easily convert a `&Vec` to `&[]`, but not the other way around.

Same suggestion applies to `run_int`.

seastar/src/config_and_start_seastar.cc
```diff
@@ -0,0 +1,110 @@
...
+        return (uint32_t)get_current_cpuset().size();
+    }
+}
+
+void set_name(const std::unique_ptr<seastar_options>& opts, const rust::Str name) {
+    opts->name = sstring(name.begin(), name.size());
+}
+
+void set_description(const std::unique_ptr<seastar_options>& opts, const rust::Str description) {
+    opts->description = std::string(description.begin(), description.size());;
```
piodul: Nit: double semicolon at the end of the line

seastar/src/config_and_start_seastar.cc
```diff
@@ -0,0 +1,110 @@
...
+    char** av = (char**) calloc(args.size() + 1, sizeof(char*));
+    if (av == nullptr) {
+        return nullptr;
+    }
+
+    for (size_t i = 0; i < args.size(); i++) {
+        const rust::String& arg = args[i];
+        av[i] = (char*) calloc(arg.size() + 1, sizeof(char));
+        if (av[i] == nullptr) {
+            return nullptr;
```
piodul, line 23: Oops, you are leaking memory here. What about the strings that you have already allocated and the `av` array?

seastar/src/config_and_start_seastar.rs
```diff
@@ -0,0 +1,293 @@
...
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_new_options_contain_default_values() {
+        let opts = Options::new();
+        assert_eq!(opts.get_name(), "App");
+        assert_eq!(opts.get_description(), "");
+        assert_eq!(opts.get_smp(), num_cpus::get() as u32);
```
piodul, line 257: Are `opts.get_smp()` and `num_cpus::get()` guaranteed to return the same on all systems? Do they return exactly the same thing?

seastar/src/config_and_start_seastar.hh
kfernandez31: Rust doesn't have access to that type directly, as it's opaque. It gets it under a `UniquePtr`, so it must transfer it through a `UniquePtr` as well.

seastar/src/config_and_start_seastar.cc
kfernandez31: Sure, I can use `operator new`. However, I'm not convinced about the second part, especially after the memory leak described in the other comment will be resolved. I tried to do it in Rust but it wasn't as pleasant.

seastar/src/config_and_start_seastar.rs
kfernandez31: Agree.

seastar/src/config_and_start_seastar.cc
kfernandez31: Must have been my Ocaml muscle memory. I just noticed, there should be `seastar::sstring`, not `std::string` too.

seastar/src/config_and_start_seastar.cc
kfernandez31: Yep, thanks.

seastar/src/config_and_start_seastar.rs
kfernandez31: Seastar:
```cpp
/// Configuration for the multicore aspect of seastar.
struct smp_options : public program_options::option_group {
    /// Number of threads (default: one per CPU).
    program_options::value<unsigned> smp;
 ...
 ```
 `num_cpus::get`: "This function will get the number of logical cores. [...]"
Due to Seastar's thread-per-core model, I should use `num_cpus::get_physical`, right?

seastar/src/config_and_start_seastar.cc
kfernandez31: Alright.

seastar/build.rs
kfernandez31: Sure. In fact, such a variable already exists. It just wasn't used here.

seastar/src/config_and_start_seastar.hh
kfernandez31: Indeed, that'd even make for shorter code.

piodul:
This PR has been superseded by https://github.com/zpp-2022-rust-seastar/seastar-rs/pull/23, closing.

PR #22, opened 2023-01-28
Spawn tasks

### LINKED ISSUE #5: Add a way to spawn tasks on the local shard in Rust, similarly to tokio::spawn

sylwiaszunejko:
Fixes: #5 
Depends on: #24, #27 

This PR introduces a way to spawn tasks on the local shard in Rust by method:
```rust
pub async fn spawn<T, Ret: 'static>(future: T) -> Option<Ret>
where
    T: Future<Output = Ret> + 'static
```
For example:
```rust
let _ = spawn(async move {
            spawn(async move {
                return 2;
            })
            .await
            .unwrap()
        })
        .await;
```

### DIFF
--- a/seastar/build.rs
+++ b/seastar/build.rs
@@ -5,9 +5,14 @@ static CXX_BRIDGES: &[&str] = &[
     "src/preempt.rs",
     "src/config_and_start_seastar.rs",
     "src/api_safety.rs",
+    "src/spawn.rs",
 ];
 
-static CXX_CPP_SOURCES: &[&str] = &["src/config_and_start_seastar.cc"];
+static CXX_CPP_SOURCES: &[&str] = &[
+    // Put all cpp source files into this list
+    "src/config_and_start_seastar.cc",
+    "src/spawn.cc",
+];
 
 fn main() {
     let seastar = pkg_config::Config::new()
--- a/seastar/src/lib.rs
+++ b/seastar/src/lib.rs
@@ -3,14 +3,13 @@
 //! Work in progress! Definitely not for use in production yet.
 
 mod api_safety;
+mod config_and_start_seastar;
 mod cxx_async_futures;
 mod cxx_async_local_future;
-
-mod config_and_start_seastar;
 mod preempt;
-
 #[cfg(test)]
 pub(crate) mod seastar_test_guard;
+mod spawn;
 
 #[cfg(test)]
 pub(crate) use seastar_test_guard::acquire_guard_for_seastar_test;
@@ -18,6 +17,7 @@ pub(crate) use seastar_test_guard::acquire_guard_for_seastar_test;
 pub use api_safety::*;
 pub use config_and_start_seastar::*;
 pub use preempt::*;
+pub use spawn::*;
 
 /// A macro intended for running asynchronous tests.
 ///
--- a/seastar/src/spawn.hh
+++ b/seastar/src/spawn.hh
@@ -1,0 +1,13 @@
+#pragma once
+
+#include "cxx-async/include/rust/cxx_async_seastar.h"
+#include "rust/cxx.h"
+#include "cxx_async_futures.hh"
+
+namespace seastar_ffi {
+namespace spawn {
+
+VoidFuture cpp_spawn(VoidFuture future);
+
+} // namespace spawn
+} // namespace seastar_ffi
--- a/seastar/src/spawn.rs
+++ b/seastar/src/spawn.rs
@@ -1,0 +1,48 @@
+use crate as seastar;
+use crate::cxx_async_local_future::IntoCxxAsyncLocalFuture;
+use core::cell::Cell;
+use ffi::*;
+use std::future::Future;
+use std::rc::Rc;
+
+#[cxx::bridge]
+mod ffi {
+    #[namespace = "seastar_ffi"]
+    unsafe extern "C++" {
+        type VoidFuture = crate::cxx_async_futures::VoidFuture;
+    }
+
+    #[namespace = "seastar_ffi::spawn"]
+    unsafe extern "C++" {
+        include!("seastar/src/spawn.hh");
+
+        fn cpp_spawn(future: VoidFuture) -> VoidFuture;
+    }
+}
+
+/// Spawns a new asynchronous task, returning an `Ret`.
+///
+/// The provided future will start running in the background immediately
+/// when `spawn` is called.
+///
+/// Spawning a task enables the task to execute concurrently to other tasks.
+///
+/// This function must be called from the context of a Seastar runtime.
+pub async fn spawn<T, Ret: 'static>(future: T) -> Ret
+where
+    T: Future<Output = Ret> + 'static,
+{
+    seastar::assert_runtime_is_running();
+
+    let x: Rc<Cell<Option<Ret>>> = Default::default();
+
+    let x_clone = x.clone();
+    match cpp_spawn(VoidFuture::infallible_local(async move {
+        x_clone.set(Some(future.await));
+    }))
+    .await
+    {
+        Ok(_) => x.take().unwrap(),
+        Err(_) => panic!(),
+    }
+}
--- a/seastar/src/spawn.rs
+++ b/seastar/src/spawn.rs
@@ -46,3 +46,44 @@ where
         Err(_) => panic!(),
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[seastar::test]
+    async fn test_empty_spawn_void() {
+        assert!(matches!(spawn(async move {}).await, ()));
+    }
+
+    #[seastar::test]
+    async fn test_chained_spawn_void() {
+        let res = spawn(async move {
+            let _ = spawn(async move {}).await;
+        })
+        .await;
+        assert!(matches!(res, ()));
+    }
+
+    #[seastar::test]
+    async fn test_spawn_int() {
+        let res = spawn(async move { 0 }).await;
+        assert!(matches!(res, 0));
+    }
+
+    #[seastar::test]
+    async fn test_two_spawn_int_and_void() {
+        let mut res = spawn(async move { 0 }).await;
+        assert!(matches!(res, 0));
+        res = spawn(async move { 1 }).await;
+        assert!(matches!(res, 1));
+        assert!(matches!(spawn(async move {}).await, ()));
+        assert!(matches!(spawn(async move {}).await, ()));
+    }
+
+    #[seastar::test]
+    async fn test_chained_spawn_int() {
+        let res = spawn(async move { spawn(async move { 2 }).await }).await;
+        assert!(matches!(res, 2));
+    }
+}
--- a/seastar/Cargo.toml
+++ b/seastar/Cargo.toml
@@ -9,6 +9,7 @@ links = "seastar"
 [dependencies]
 cxx = "1"
 cxx-async = { git = "https://github.com/kfernandez31/cxx-async", branch = "seastar" }
+futures = "0.3.25"
 pin-project = "1"
 seastar-macros = { path = "../seastar-macros" }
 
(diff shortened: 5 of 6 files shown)


seastar/src/spawn.rs
```diff
@@ -41,3 +41,47 @@ where
...
+    async fn test_spawn_int() {
+        let res = spawn(async move {
+            return 0;
+        })
+        .await;
+        assert!(matches!(res, Some(0)));
+    }
+
+    #[seastar::test]
+    async fn test_two_spawn_int_and_void() {
```
kfernandez31, line 75: I think we can replace this test with a "`test_spawn_void`" to avoid redundancy.

seastar/src/spawn.rs
```diff
@@ -1,53 +1,43 @@
...
         include!("seastar/src/spawn.hh");
 
-        fn cpp_spawn_void(future: VoidFuture) -> VoidFuture;
-        fn cpp_spawn_int(future: IntFuture) -> IntFuture;
+        fn cpp_spawn(future: VoidFuture) -> VoidFuture;
     }
 }
 
-/// Spawns a new asynchronous task, returning a `VoidFuture`.
+/// Spawns a new asynchronous task, returning a `Option<Ret>`.
```
kfernandez31: Nit: "a" -> "an" :p

kfernandez31 (changes requested):

seastar/src/spawn.rs
sylwiaszunejko: Do you mean I should delete this test? I wrote it because there was a moment where a single spawn worked, but multiple spawns in a row didn't.

seastar/src/spawn.rs
```diff
@@ -0,0 +1,43 @@
...
+
+/// Spawns a new asynchronous task, returning an `Option<Ret>`.
+///
+/// The provided future will start running in the background immediately
+/// when `spawn` is called.
+///
+/// Spawning a task enables the task to execute concurrently to other tasks.
+///
+/// This function must be called from the context of a Seastar runtime.
+pub async fn spawn<T, Ret: 'static>(future: T) -> Option<Ret>
```
piodul: Why does it return an option? I think it should just return `cxx_async::CxxAsyncResult<Ret>`, like in submit_to.

seastar/src/spawn.rs
```diff
@@ -0,0 +1,43 @@
...
+    T: Future<Output = Ret> + 'static,
+{
+    let x: Rc<Cell<Option<Ret>>> = Default::default();
+    let xcloned = x.clone();
+    cpp_spawn(VoidFuture::infallible_local(async move {
+        xcloned.set(Some(future.await));
+    }))
+    .await
+    .ok()?;
+    Rc::try_unwrap(x).ok().unwrap().into_inner()
```
piodul: This line is very strange. Why `Rc::try_unwrap`?

I would imagine something like this:

```rust
x.take().unwrap()
```

seastar/src/lib.rs
```diff
@@ -8,16 +8,17 @@ mod cxx_async_local_future;
 
 mod config_and_start_seastar;
 mod preempt;
-
```
piodul, line 11: Nit: unnecessary whitespace change

seastar/src/spawn.rs
```diff
@@ -32,6 +32,8 @@ pub async fn spawn<T, Ret: 'static>(future: T) -> Option<Ret>
 where
     T: Future<Output = Ret> + 'static,
 {
+    seastar::assert_runtime_is_running!();
```
piodul: `Ensure safety of spawn`

Please squash this commit with `Add functions to spawn future on the local shard`.

piodul:
Please rebase and drop the commit [Add macro #[seastar::test]](https://github.com/zpp-2022-rust-seastar/seastar-rs/pull/22/commits/b71681613620e47f4460f8f147deba51299be4f8) (the PR for this was merged), as there are conflicts.

sylwiaszunejko:
> Please rebase and drop the commit [Add macro #[seastar::test]](https://github.com/zpp-2022-rust-seastar/seastar-rs/pull/22/commits/b71681613620e47f4460f8f147deba51299be4f8) (the PR for this was merged), as there are conflicts.

done

piodul:
We merged the "API safety" PR, so there are conflicts again... please do the same as before - drop [Add module for checking whether current thread is in a Seastar runtime](https://github.com/zpp-2022-rust-seastar/seastar-rs/pull/22/commits/a664964a26d33b3a5359ccac39cadb850b19a214) and rebase on current main.

Apart from that, I think it looks good to be merged.

seastar/build.rs
```diff
@@ -4,9 +4,11 @@ static CXX_BRIDGES: &[&str] = &[
     // Put all files that contain a cxx::bridge into this list
     "src/preempt.rs",
     "src/config_and_start_seastar.rs",
+    "src/api_safety.rs",
+    "src/spawn.rs",
 ];
 
-static CXX_CPP_SOURCES: &[&str] = &["src/config_and_start_seastar.cc"];
+static CXX_CPP_SOURCES: &[&str] = &["src/config_and_start_seastar.cc", "src/spawn.cc"];
```
piodul: One more thing. As each PR will add files to this variable, it would be well to have each file put in a separate line so that we reduce the risk of a merge conflict.

I think a trick used in `CXX_BRIDGES` will prevent rustfmt from putting everything back into a single line: just add a comment similar to the `Put all files...`. Please add it.

seastar/src/spawn.hh
```diff
@@ -0,0 +1,13 @@
...
+#include "cxx-async/include/rust/cxx_async_seastar.h"
+#include "rust/cxx.h"
+#include "cxx_async_futures.hh"
+
+namespace seastar_ffi {
+namespace spawn {
+
+VoidFuture cpp_spawn(VoidFuture future);
+
+}
```
kfernandez31: Nit: no `// namespace <namespace name>` comments (it's a good practice to have them)

seastar/src/spawn.rs
```diff
@@ -0,0 +1,47 @@
...
+/// This function must be called from the context of a Seastar runtime.
+pub async fn spawn<T, Ret: 'static>(future: T) -> cxx_async::CxxAsyncResult<Ret>
+where
+    T: Future<Output = Ret> + 'static,
+{
+    seastar::assert_runtime_is_running!();
+
+    let x: Rc<Cell<Option<Ret>>> = Default::default();
+    let xcloned = x.clone();
+    match cpp_spawn(VoidFuture::infallible_local(async move {
```
kfernandez31, line 40: This can be shortened:
```rs
    seastar::assert_runtime_is_running!();

    let x: Rc<Cell<Option<Ret>>> = Default::default();

    let x_clone = x.clone();
    cpp_spawn(VoidFuture::infallible_local(async move {
        x_clone.set(Some(future.await));
    }))
    .await?;

    Ok(x.take().unwrap())
    ```

File: seastar/src/spawn.rs (hunk 1 above)
Comment (kfernandez31) [reply seastar/src/spawn.rs:75]: I see. What bugs me is the asymmetry of the `spawn_int` and `spawn_void` calls. I'd add a second `spawn_int` call in this test and then either keep it as is or split it into two.

Comment (kfernandez31) [reply seastar/src/spawn.rs:75]: You can also think of testing those tests' logic with the `join!` macro - it'd make them more general over the order of the spawns (of course, in these tests, the tasks are kind of identical, so it doesn't matter, but that doesn't have to be the case).

Comment (sylwiaszunejko) [reply seastar/src/spawn.rs:75]: But there are two `spawn_int` calls. I still don't get what the problem is.

File: seastar/src/spawn.rs [hunk 10]
```diff
@@ -0,0 +1,87 @@
...
+
+    let x: Rc<Cell<Option<Ret>>> = Default::default();
+
+    let x_clone = x.clone();
+    cpp_spawn(VoidFuture::infallible_local(async move {
+        x_clone.set(Some(future.await));
+    }))
+    .await?;
+
+    Ok(x.take().unwrap())
```
Lorak-mmk: I'm not yet completely convinced about usage of `Rc<Cell<Option<T>>>` - see comment https://github.com/zpp-2022-rust-seastar/seastar-rs/pull/26#discussion_r1125474021 as it applies here too.

sylwiaszunejko: Do you suggest using oneshot::channel? Previously I think @piodul said that it is overkill because the channel will be unnecessarily thread-safe.

piodul: Based on the precedent made here: https://github.com/zpp-2022-rust-seastar/seastar-rs/pull/26#discussion_r1125669863 I think we can keep using `Rc<Cell<Option<T>>>` here as well.

A thread-safe channel is an overkill IMO. The `Rc/Cell/Option` stuff is hidden behind a nice abstraction so this should be fine.

piodul: If this pattern repeats frequently then I'm fine - in the future - with either creating or using (from an external crate) a nicer abstraction for this with desired semantics and performance characteristics.

seastar/src/spawn.rs
```diff
@@ -0,0 +1,87 @@
...
+
+/// Spawns a new asynchronous task, returning an `cxx_async::CxxAsyncResult<Ret>`.
+///
+/// The provided future will start running in the background immediately
+/// when `spawn` is called.
+///
+/// Spawning a task enables the task to execute concurrently to other tasks.
+///
+/// This function must be called from the context of a Seastar runtime.
+pub async fn spawn<T, Ret: 'static>(future: T) -> cxx_async::CxxAsyncResult<Ret>
```
piodul: One more thing, related to the conversation on the chat: I don't think it actually makes sense for this function return a Result - it should just return Ret. The only place where a failed result can be generated is `cpp_spawn`, which indicates an exception on the c++ side. I can't imagine any other failure apart from an allocation failure, and Rust just panics/aborts on allocation failure, so I think it makes sense to do the same here - and it simplifies the interface.

piodul (commented):
Looks good to me to merge after changing the signature to return `Ret`, not `Result<Ret, _>`.

seastar/src/spawn.rs
Lorak-mmk: If we want to use this with Scylla, then don't we need to handle allocation failures?

piodul: Not necessarily. Seastar itself considers some allocations to be irrecoverable. Grep for [scoped_critical_alloc_section](https://github.com/scylladb/seastar/blob/799a9aacb53db4f13c35caecf5071cf078b37159/include/seastar/util/critical_alloc_section.hh#L46-L68) in the Seastar code and see for yourself, it is frequently used in future-related primitives.

Besides that, Rust is ill-prepared to handle allocation failures at this moment of time, unfortunately. Some primitives that allocate (Box, Vec) do have methods that allow handling such failures, but they are opt-in and everything else just aborts - not even panic. We will just have to be super careful not to allocate large chunks of data and use fallible allocation APIs for user data.

Finally, remember that aborting is considered to be safe behavior from Rust's POV.

piodul (approved):

PR #23, opened 2023-01-28
Add module for configuring and starting seastar

piodul:
This PR introduces the configure_and_start_seastar module. As the name suggests, it provides functionality necessary to configure and start the seastar runtime.

It contains definitions of the following types:

- `Options` - wrapper over `seastar::app_template::options`. Provides configuration for the seastar runtime.
- `AppTemplate` - wrapper over `seastar::app_template`. Used to start the seastar runtime.

Example:
```rust
let mut app = AppTemplate::default();
let args = vec!["test"];
let fut = async { Ok(42) };
assert_eq!(app.run_int(&args[..], fut), 42);
```

Heavily based on work done in: https://github.com/zpp-2022-rust-seastar/seastar-rs/pull/10 by @kfernandez31, with the help from @sylwiaszunejko and @patjed41.

### DIFF
--- a/seastar/src/config_and_start_seastar.cc
+++ b/seastar/src/config_and_start_seastar.cc
@@ -1,0 +1,57 @@
+#include "config_and_start_seastar.hh"
+
+namespace seastar_ffi {
+namespace config_and_start_seastar {
+
+std::unique_ptr<seastar_options> new_options() {
+    return std::make_unique<seastar_options>();
+}
+
+rust::Str get_name(const seastar_options& opts) {
+    return rust::Str(opts.name.begin(), opts.name.size());
+}
+
+rust::Str get_description(const seastar_options& opts) {
+    return rust::Str(opts.description.begin(), opts.description.size());
+}
+
+uint32_t get_smp(const seastar_options& opts) {
+    if (opts.smp_opts.smp) {
+        return (uint32_t)opts.smp_opts.smp.get_value();
+    } else {
+        return (uint32_t)seastar::get_current_cpuset().size();
+    }
+}
+
+void set_name(seastar_options& opts, const rust::Str name) {
+    opts.name = seastar::sstring(name.begin(), name.size());
+}
+
+void set_description(seastar_options& opts, const rust::Str description) {
+    opts.description = seastar::sstring(description.begin(), description.size());
+}
+
+void set_smp(seastar_options& opts, const uint32_t smp) {
+    opts.smp_opts.smp.set_value((unsigned)smp);
+}
+
+std::unique_ptr<app_template> new_app_template_from_options(seastar_options& opts) {
+    return std::make_unique<app_template>(std::move(opts));
+}
+
+int32_t run_void(app_template& app, int32_t argc, char** argv, VoidFuture fut) {
+    int32_t exit_value = app.run((int)argc, argv, [&]() -> seastar::future<> {
+        co_await std::move(fut);
+    });
+    return exit_value;
+}
+
+int32_t run_int(app_template& app, int32_t argc, char** argv, IntFuture fut) {
+    int32_t exit_value = app.run((int)argc, argv, [&]() -> seastar::future<int> {
+        co_return co_await std::move(fut);
+    });
+    return exit_value;
+}
+
+} // namespace config_and_start_seastar
+} // namespace seastar
--- a/seastar/src/config_and_start_seastar.rs
+++ b/seastar/src/config_and_start_seastar.rs
@@ -361,6 +361,7 @@ mod tests {
     #[test]
     fn test_run_int() {
         thread::spawn(|| {
+            let _guard = crate::acquire_guard_for_seastar_test();
             let mut app = AppTemplate::default();
             let args = vec!["test"];
             let fut = async { Ok(42) };
@@ -373,6 +374,7 @@ mod tests {
     #[test]
     fn test_run_void() {
         thread::spawn(|| {
+            let _guard = crate::acquire_guard_for_seastar_test();
             let mut app = AppTemplate::default();
             let args = vec!["test"];
             let fut = async { Ok(()) };
--- a/seastar/src/config_and_start_seastar.rs
+++ b/seastar/src/config_and_start_seastar.rs
@@ -1,0 +1,398 @@
+use std::{
+    ffi::{c_char, CString, OsString},
+    future::Future,
+};
+
+use cxx::UniquePtr;
+use ffi::*;
+
+use crate::cxx_async_local_future::IntoCxxAsyncLocalFuture;
+
+#[cxx::bridge]
+mod ffi {
+    #[namespace = "seastar_ffi"]
+    unsafe extern "C++" {
+        type VoidFuture = crate::cxx_async_futures::VoidFuture;
+        type IntFuture = crate::cxx_async_futures::IntFuture;
+    }
+
+    #[namespace = "seastar_ffi::config_and_start_seastar"]
+    unsafe extern "C++" {
+        include!("seastar/src/config_and_start_seastar.hh");
+
+        type seastar_options;
+        type app_template;
+
+        // Returns a pointer to default `seastar_options`
+        fn new_options() -> UniquePtr<seastar_options>;
+        // Getters
+        fn get_name(opts: &seastar_options) -> &str;
+        fn get_description(opts: &seastar_options) -> &str;
+        fn get_smp(opts: &seastar_options) -> u32;
+        // Setters
+        fn set_name(opts: Pin<&mut seastar_options>, name: &str);
+        fn set_description(opts: Pin<&mut seastar_options>, description: &str);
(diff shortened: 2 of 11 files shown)


piodul:
@cvybhu review ping

seastar/src/config_and_start_seastar.cc
```diff
@@ -0,0 +1,105 @@
...
+        av[i] = new(std::nothrow) char[arg.size() + 1];
+        if (av[i] == nullptr) {
+            for (size_t j = 0; j < i; j++) {
+                delete av[j];
+            }
+            delete[] av;
+            return nullptr;
+        }
+        strncpy(av[i], arg.data(), arg.size());
+    }
```
cvybhu: I think the strings in `av[i]` might not be null-terminated properly. `strncpy` copies  `n` bytes from the slice, but rust str slices don't contain the null byte.
Adding `()` would zero-initialize the memory:
```c++
av[i] = new(std::nothrow) char[arg.size() + 1]();
```

There might also be utf8 characters that contain a null byte, `strncpy` will stop in the middle of the word then. `memcpy` wouldn't have this issue.

```c
// man strncpy:
// A simple implementation of strncpy() might be:
char* strncpy(char *dest, const char *src, size_t n) {
    size_t i;

    for (i = 0; i < n && src[i] != '\0'; i++)
        dest[i] = src[i];
    for ( ; i < n; i++)
        dest[i] = '\0';

    return dest;
}       
```

seastar/src/config_and_start_seastar.rs
```diff
@@ -0,0 +1,339 @@
...
+    /// let app = AppTemplate::new_from_options(Options::default());
+    /// ```
+    pub fn new_from_options(mut opts: Options) -> Self {
+        AppTemplate {
+            app: new_app_template_from_options(opts.opts.pin_mut()),
+        }
+    }
+
+    /// Runs an app with a void callback (the output of which is always 0) and program arguments (argv).
+    /// The app is run on a separate thread.
```
cvybhu: When I read this my understanding was that `run_void` would start a new thread for me and run the app there. 
Let's change it to something like:
`Each call to run_void must be done on a new thread`, or something like that (?)

seastar/src/config_and_start_seastar.rs
```diff
@@ -0,0 +1,339 @@
...
+    /// let mut app = AppTemplate::default();
+    /// let args = vec!["hello"];
+    ///
+    /// assert_eq!(app.run_int(&args[..], fut), 42);
+    /// ```
+    pub fn run_int<I, Arg>(
+        &mut self,
+        args: I,
+        fut: impl Future<Output = cxx_async::CxxAsyncResult<i32>> + 'static,
+    ) -> i32
```
cvybhu, line 270: It would be nicer to have `Result<>` instead of returning 1 on allocation error.

seastar/src/seastar_test_guard.rs
```diff
@@ -0,0 +1,21 @@
...
+/// multiple times in a sequence, it doesn't like (and doesn't detect)
+/// when multiple runtimes are trying to run in parallel. `cargo test` runs
+/// tests in parallel by default, so it's easy to write tests which will
+/// make seastar angry (the anger manifests itself in form of segfaults).
+///
+/// The mutex should be taken by all tests that create a seastar runtime
+/// and held until the test finishes.
+pub(crate) fn acquire_guard_for_seastar_test() -> RunningTestWithSeastarGuard {
+    let guard = RUNNING_TEST_WITH_SEASTAR.lock().unwrap();
+    RunningTestWithSeastarGuard(guard)
```
cvybhu: I wonder if a panic in test will poison the mutex. We could catch the panic or something (?) I guess it's not that important.

cvybhu (changes requested):

seastar/src/config_and_start_seastar.rs
cvybhu: Or maybe just panic!() since we're writing rust

seastar/src/config_and_start_seastar.rs
```diff
@@ -0,0 +1,341 @@
...
+    pub fn get_name(&self) -> &str {
+        get_name(&self.opts)
+    }
+
+    /// Gets the `Options`' description.
+    ///
+    /// # Examples
+    ///
+    /// ```rust
+    /// use seastar::{AppTemplate, Options};
```
kfernandez31: Unused import `AppTemplate` here and in other docs.

seastar/src/config_and_start_seastar.rs
piodul: I think a panic makes sense - it's consistent with how Rust handles allocation errors. I'll probably convert  `args_as_ptr` to Rust anyway, so we will get this behavior automatically.

seastar/src/config_and_start_seastar.cc
piodul: Good catch. I'll try to RIIR this function - the standard library conversion routines should be aware of those issues.

piodul: Oh, how convenient, cxx has an example which, more or less, does what we want to do here: https://cxx.rs/binding/rawptr.html

seastar/src/seastar_test_guard.rs
piodul: Yes, it will poison the mutex. If other tests try to acquire the mutex, they will get an error.

I'm not sure if it matters at the moment. If a panic occurs and tries to cross the boundary through cxx, we will get an abort. However, if we solve this problem in the future then, after a test panics, the seastar runtime will have been properly stopped. In such case it will make sense to ignore the poisoning - so I'll modify the current code to ignore the poison error.

piodul:
v2:
* Changed run_void/run_int to accept an iterator over items convertible to OsString (so that it works with std::env::args_os()),
* Now, Rust converts the OS arguments and passes them to C++ as a `char**`,
* Poisoning is ignored in `acquire_guard_for_seastar_test`,
* Various doc fixes.

seastar/src/cxx_async_futures.hh
```diff
@@ -0,0 +1,7 @@
+#include "rust/cxx.h"
+#include "cxx-async/include/rust/cxx_async.h"
+#include "cxx-async/include/rust/cxx_async_seastar.h"
+#include <seastar/core/coroutine.hh>
+
+CXXASYNC_DEFINE_FUTURE(void, seastar_ffi, config_and_start_seastar, VoidFuture);
```
kfernandez31: I think we'd rather have any `cxx-async` futures to be declared directly under the `seastar_ffi` namespace, as they won't be unique to the context of `config_and_start_seastar.*` but will be used in other places as well (`seastar_ffi::timer` and `seastar_ffi::spawn` to name a few). Of course they can still be used there, but the second namespace is slightly misleading.

seastar/src/config_and_start_seastar.rs
```diff
@@ -0,0 +1,390 @@
...
+            let v = os.into().into_vec();
+            CString::new(v).unwrap_or_else(|err| {
+                let nul_position = err.nul_position();
+                let mut v = err.into_vec();
+                v.truncate(nul_position);
+                CString::new(v).unwrap()
+            })
+        })
+        .collect()
+}
```
cvybhu, line 313: I read on SO that `char** argv[]` has to be null terminated (have `argv[argc] = NULL`)
https://stackoverflow.com/a/35335248

With Vec<CString> this wouldn't be the case.

seastar/src/cxx_async_futures.hh
patjed41: I think that the easiest way to achieve this is by changing
```rust
CXXASYNC_DEFINE_FUTURE(void, seastar_ffi, config_and_start_seastar, VoidFuture);
```
to
```rust
CXXASYNC_DEFINE_FUTURE(void, seastar_ffi, VoidFuture);
```
and by defining `VoidFuture` in `extern "C++"` block like that
```rust
#[namespace = "seastar_ffi"]
type VoidFuture = crate::cxx_async_futures::VoidFuture;
```
in every `ffi` module that uses `VoidFuture`. At least this what I've done to be able to use `seastar_ffi::<some_other_namespace>` instead of `seastar_ffi::config_and_start_seastar` in another task.

piodul: Yes, I agree that this is a good idea. Changed.

seastar/src/config_and_start_seastar.rs
piodul: I didn't know about that, thanks. This makes `argc` a bit redundant, but if the null pointer in the end is required - so be it.

piodul:
v2:
- Moved the cxx-async future definitions to the `seastar_ffi` module
- Added the mandatory null pointer at the end of the `args` vector

piodul:
@cvybhu ping

cvybhu (approved):

PR #24, opened 2023-02-11
Test macro

### LINKED ISSUE #2: Create a macro like tokio::test

grzenow4:
Fixes: #2
Depends on #1

This PR introduces a procedural macro like `tokio::test` which allows writing tests in Rust that run on Seastar runtime. For example:
```rust
#[seastar::test]
async fn test_frobnication() {
    // You can use seastar functions here without any problems, and await on them
    do_frobnicate().await;
}
```
With the help from @kfernandez31

### DIFF
--- a/seastar-macros/Cargo.toml
+++ b/seastar-macros/Cargo.toml
@@ -1,0 +1,14 @@
+[package]
+name = "seastar-macros"
+version = "0.1.0"
+edition = "2021"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[lib]
+proc-macro = true
+
+[dependencies]
+syn = "1.0"
+quote = "1.0"
+proc-macro2 = "1.0"
--- a/seastar-macros/src/lib.rs
+++ b/seastar-macros/src/lib.rs
@@ -1,0 +1,44 @@
+#[proc_macro_attribute]
+pub fn test(
+    args: proc_macro::TokenStream,
+    item: proc_macro::TokenStream,
+) -> proc_macro::TokenStream {
+    let args = syn::parse_macro_input!(args as proc_macro2::TokenStream);
+    let input = syn::parse_macro_input!(item as syn::ItemFn);
+
+    let name = &input.sig.ident;
+    let body = &input.block;
+    let attrs = &input.attrs;
+
+    if input.sig.asyncness.is_none() {
+        let msg = "the async keyword is missing from the function declaration";
+        return syn::Error::new_spanned(input.sig, msg)
+            .to_compile_error()
+            .into();
+    }
+
+    if !args.is_empty() {
+        let msg = "arguments for #[seastar::test] are not supported yet";
+        return syn::Error::new_spanned(args, msg).to_compile_error().into();
+    }
+
+    let output = quote::quote! {
+        #[test]
+        #(#attrs)*
+        fn #name() {
+            std::thread::spawn(|| {
+                let _guard = seastar::acquire_guard_for_seastar_test();
+                let mut app = seastar::AppTemplate::default();
+                let fut = async {
+                    #body
+                    Ok(())
+                };
+                app.run_void(std::env::args().take(1), fut);
+            })
+            .join()
+            .unwrap();
+        }
+    };
+
+    output.into()
+}
--- a/seastar/src/lib.rs
+++ b/seastar/src/lib.rs
@@ -16,3 +16,19 @@ pub(crate) use seastar_test_guard::acquire_guard_for_seastar_test;
 
 pub use config_and_start_seastar::*;
 pub use preempt::*;
+
+/// A macro intended for running asynchronous tests.
+///
+/// Tests are spawned in a separate thread.
+/// This is done to ensure thread_local cleanup between them
+/// (at the time of writing, Seastar doesn't do it itself).
+///
+/// # Usage
+///
+/// ```rust
+/// #[seastar::test]
+/// async fn my_test() {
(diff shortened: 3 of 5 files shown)


piodul (changes requested):
The commits in the PR should be cleaned up before we can continue with review. It should be possible to understand what this PR achieves and how by going through it commit-by-commit and reading each commit's message and diff.

I already provided some tips and reasons for this before on another PR, here: https://github.com/zpp-2022-rust-seastar/zpp2022-seastar-rust-integration-playground/pull/2#issuecomment-1336428751

I'll sum up the relevant rules here:

- Don't submit PRs with merge commits. If you want to base off some other branch, start your branch from it and rebase on top of it if that branch gets updated. It's also good to avoid merge commits during development as they are hard to get rid of later when cleaning up.
- Each commit should have a clear message and focus on doing one thing - e.g. perform some refactor, add some features. The commit message must be clear and the commit diff readable.
- The project must compile and pass tests/linting after each commit.

There are some merge commits on the PR's branch right now, as far as I know they can be problematic if you try to do a rebase such branch. I don't have much experience here, but you can try some approaches:
- `git rebase -i piodul/configure-runtime-redux test_macro` should get rid of the merge commits, but there might be some conflicts that are difficult to resolve.
- You can try creating a new branch from `piodul/configure-runtime-redux`, then cherry-pick some commits from the old branch, then move the old branch to point at the new one via `git branch -f test_macro name-of-the-new-branch`
- If the above won't work, there is a nuclear option: creating a single commit based on a diff between branches. You can try the process described here: https://stackoverflow.com/a/17324759 , and then you can continue with splitting up the commit into smaller parts.

seastar-macros/Cargo.toml
```diff
@@ -0,0 +1,16 @@
+[package]
+name = "seastar-macros"
+version = "0.1.0"
+edition = "2021"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[lib]
+path = "src/lib.rs"
```
Lorak-mmk: I think this is the default value, and so doesn't need to be specified.

seastar-macros/Cargo.toml
```diff
@@ -0,0 +1,16 @@
...
+proc-macro = true
+
+[dependencies]
+syn = "1.0"
+quote = "1.0"
+proc-macro2 = "1.0"
+futures = "0.3"
```
Lorak-mmk: nit: no newline at the end of file

seastar-macros/src/lib.rs
```diff
@@ -0,0 +1,45 @@
...
+    if input.sig.asyncness.is_none() {
+        let msg = "the async keyword is missing from the function declaration";
+        return syn::Error::new_spanned(input.sig.fn_token, msg)
+            .to_compile_error()
+            .into();
+    }
+
+    if !args.is_empty() {
+        let msg = "arguments for #[seastar::test] are not supported yet";
+        return syn::Error::new_spanned(input.sig.inputs, msg)
```
Lorak-mmk: I've never written a proc macro so take this comment with a grain of salt.
Why use `input.sig.inputs` instead of `args`? The formar also has `ToTokens` implementation, so you should be able to use it, and `args` is the actual place where the error is, not function signature.

seastar-macros/src/lib.rs
```diff
@@ -0,0 +1,45 @@
...
+) -> proc_macro::TokenStream {
+    let input = syn::parse_macro_input!(item as syn::ItemFn);
+
+    let name = &input.sig.ident;
+    let body = &input.block;
+    let attrs = &input.attrs;
+
+    if input.sig.asyncness.is_none() {
+        let msg = "the async keyword is missing from the function declaration";
+        return syn::Error::new_spanned(input.sig.fn_token, msg)
```
Lorak-mmk: I've never written a proc macro so take this comment with a grain of salt.
What is `fn_token`, what does it span? Function name? signature? whole function?
`input.sig` should be possible to use it, and it sounds like it should span the signature (I may be mistaken of course), which is where the error should be shown.

Lorak-mmk (changes requested):

seastar/src/lib.rs
```diff
@@ -16,3 +16,18 @@ pub(crate) use seastar_test_guard::acquire_guard_for_seastar_test;
 
 pub use config_and_start_seastar::*;
 pub use preempt::*;
+
+/// A macro intended for running asynchronous tests.
```
piodul, line 20: Nit: put an empty comment line after this sentence. `cargo doc` uses the first block of text as a short summary of the item:

```rust
/// A macro intended for running asynchronous tests.
///
/// Tests are spawned in a separate thread.
/// This is done to ensure thread_local cleanup between them
/// (at the time of writing, Seastar doesn't do it itself).
```

piodul:
@grzenow4 It looks like there are some unresolved comments from @Lorak-mmk and me. Please fix so that we can continue and unblock other tasks.

piodul (approved):

Lorak-mmk (approved):

PR #25, opened 2023-02-11
Main macro

### LINKED ISSUE #16: Create a macro like tokio::main
[`tokio::main`](https://docs.rs/tokio/latest/tokio/attr.main.html) is a macro that takes care of starting the runtime so that the user can just write `async fn main` and use runtime functionality immediately.

We would like to have something similar - `seastar::main`.
It would be nice to be able to pass some runtime configuration options to the macro.

This issue doesn't have a very high priority, for now we can just create the runtime manually, it's more important to get all the functionality working first.

kfernandez31:
Fixes: #16
Depends: #23 

This PR introduces the `#[seastar::main]` macro. It works exactly as described in #16 .
A minimal example of usage:
```rs
#[seastar::main]
async fn main() {
    println!("Hello, world!);
}
```
What the macro does under the hood is that it calls `run_void` on an `AppTemplate` instance with default `Options`. We *may* offer some room for configuration in the future, make the macro accept options as args, but for now this is enough.


piodul (changes requested):
See my comments here: https://github.com/zpp-2022-rust-seastar/seastar-rs/pull/24#pullrequestreview-1294643460

It looks like this PR is based off the branch for `seastar::test`, but doesn't use the functionality introduced there in any way. A PR should not include unrelated changes, please remove those commits from the branch for this PR.

PR #26, opened 2023-02-12
#[seastar::main] macro

### LINKED ISSUE #16: Create a macro like tokio::main

kfernandez31:
An improvement over #25 (still learning how to do proper PRs).
Fixes: #16
Depends: #23

This PR introduces the `#[seastar::main]` macro. It works exactly as described in #16 .
A minimal example of usage:
```rs
#[seastar::main]
async fn main() {
    hello_world().await;
}
```
What the macro does under the hood is that it calls `run_void` on an `AppTemplate` instance with default `Options`. We may offer some room for configuration in the future, and make the macro accept options as args, but for now, this is enough.

### DIFF
--- a/seastar-macros/src/lib.rs
+++ b/seastar-macros/src/lib.rs
@@ -42,3 +42,70 @@ pub fn test(
 
     output.into()
 }
+
+#[proc_macro_attribute]
+#[cfg(not(test))] // Work around for rust-lang/rust#62127
+pub fn main(
+    args: proc_macro::TokenStream,
+    item: proc_macro::TokenStream,
+) -> proc_macro::TokenStream {
+    let args = syn::parse_macro_input!(args as proc_macro2::TokenStream);
+    let input = syn::parse_macro_input!(item as syn::ItemFn);
+
+    let output = &input.sig.output;
+    let inputs = &input.sig.inputs;
+    let name = &input.sig.ident;
+    let body = &input.block;
+    let attrs = &input.attrs;
+
+    if input.sig.asyncness.is_none() {
+        let msg = "the async keyword is missing from the function declaration";
+        return syn::Error::new_spanned(input.sig, msg)
+            .to_compile_error()
+            .into();
+    }
+
+    if name != "main" {
+        let msg = "only the main function is allowed to use #[seastar::main]";
+        return syn::Error::new_spanned(input.sig.ident, msg)
+            .to_compile_error()
+            .into();
+    }
+
+    if !inputs.is_empty() {
+        let msg = "the main function cannot accept arguments";
+        return syn::Error::new_spanned(input.sig, msg)
+            .to_compile_error()
+            .into();
+    }
+
+    if !args.is_empty() {
+        let msg = "arguments for #[seastar::test] are not supported yet";
+        return syn::Error::new_spanned(args, msg).to_compile_error().into();
+    }
+
+    let ret_type = match output {
+        syn::ReturnType::Type(_, ty) => ty.clone(),
+        syn::ReturnType::Default => syn::parse_quote! { () },
+    };
+
+    let output = quote::quote! {
+        #(#attrs)*
+        fn main() #output {
+            let ret_holder : std::rc::Rc<std::cell::Cell<Option<#ret_type>>> = Default::default();
+
+            let ret_holder_clone = ret_holder.clone();
+            let fut = async move {
+                let ret = #body;
+                ret_holder_clone.set(Some(ret));
+                Ok(())
+            };
+
+            let mut app = seastar::AppTemplate::default();
+            app.run_void(std::env::args(), fut);
+            ret_holder.take().unwrap()
+        }
+    };
+
+    output.into()
+}
--- a/seastar/src/lib.rs
+++ b/seastar/src/lib.rs
@@ -34,3 +34,25 @@ pub use preempt::*;
 /// }
 /// ```
 pub use seastar_macros::test;
+
+/// A macro that runs the `main` function's contents
+/// in Seastar's runtime.
+///
+/// **Only** `main` is allowed to use this macro -
+/// Seastar apps may only use one `AppTemplate`
+/// instance at a time.
+///
+/// # Options
+///
+/// Currently, passing options is not supported.
+/// The app is simply run with default parameters.
+///
+/// # Usage
+///
+/// ```rust
+/// #[seastar::main]
+/// async fn main() {
+///     println!("Hello, world!");
+/// }
+/// ```
+pub use seastar_macros::main;
--- a/seastar/examples/main_macro.rs
+++ b/seastar/examples/main_macro.rs
@@ -1,0 +1,8 @@
+async fn frobnicate() -> i32 {
+    42
+}
+
+#[seastar::main]
+async fn main() {
+    println!("Frobnication result: {}", frobnicate().await);
+}


seastar-macros/Cargo.toml
```diff
@@ -0,0 +1,15 @@
+[package]
+name = "seastar-macros"
+version = "0.1.0"
+edition = "2021"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[lib]
+path = "src/lib.rs"
```
Lorak-mmk: I think this is the default value, and so doesn't need to be specified.

seastar-macros/src/lib.rs
```diff
@@ -0,0 +1,56 @@
...
+    if !inputs.is_empty() {
+        let msg = "the main function cannot accept arguments";
+        return syn::Error::new_spanned(input.sig.inputs, msg)
+            .to_compile_error()
+            .into();
+    }
+
+    if !args.is_empty() {
+        let msg = "arguments for #[seastar::main] are not supported yet";
+        return syn::Error::new_spanned(input.sig.inputs, msg)
```
Lorak-mmk: I've never written a proc macro so take this comment with a grain of salt.
Why use `input.sig.inputs` instead of `args`? The formar also has `ToTokens` implementation, so you should be able to use it, and `args` is the actual place where the error is, not function signature.

seastar-macros/src/lib.rs
```diff
@@ -0,0 +1,56 @@
...
+    let args = syn::parse_macro_input!(args as syn::AttributeArgs);
+
+    let inputs = &input.sig.inputs;
+    let name = &input.sig.ident;
+    let body = &input.block;
+    let attrs = &input.attrs;
+
+    if input.sig.asyncness.is_none() {
+        let msg = "the async keyword is missing from the function declaration";
+        return syn::Error::new_spanned(input.sig.fn_token, msg)
```
Lorak-mmk: I've never written a proc macro so take this comment with a grain of salt.
What is `fn_token`, what does it span? Function name? signature? whole function?
`input.sig` should be possible to use it, and it sounds like it should span the signature (I may be mistaken of course), which is where the error should be shown.

seastar-macros/src/lib.rs
```diff
@@ -0,0 +1,56 @@
...
+            .to_compile_error()
+            .into();
+    }
+
+    if name != "main" {
+        let msg = "only the main function is allowed to use #[seastar::main]";
+        return syn::Error::new_spanned(input.sig.inputs, msg)
```
Lorak-mmk: Similarly to other comments: why not `input.sig.ident`?

Lorak-mmk (changes requested):

seastar/src/lib.rs
```diff
@@ -16,3 +16,23 @@ pub(crate) use seastar_test_guard::acquire_guard_for_seastar_test;
 
 pub use config_and_start_seastar::*;
 pub use preempt::*;
+
+/// A macro that runs the `main` function's contents
+/// in Seastar's runtime. **Only** `main` is allowed
+/// to use this macro - Seastar apps may only use
+/// one `AppTemplate` instance at a time.
```
piodul: The docstring should start with a block that shortly describes given item, then an empty comment line should follow, and then the rest of the docstring. `cargo doc` uses the first block as a summary of given item. This summary is e.g. displayed on the doc page of the encompassing module, on the list of module's items.

So, like this:

```rust
/// A macro that runs the `main` function's contents
/// in Seastar's runtime.
/// **Only** `main` is allowed to use this macro -
/// Seastar apps may only use one `AppTemplate`
/// instance at a time.
```

seastar-macros/src/lib.rs
```diff
@@ -0,0 +1,56 @@
...
+    if !args.is_empty() {
+        let msg = "arguments for #[seastar::main] are not supported yet";
+        return syn::Error::new_spanned(input.sig.inputs, msg)
+            .to_compile_error()
+            .into();
+    }
+
+    let output = quote::quote! {
+        #(#attrs)*
+        fn main() {
```
piodul: It would be really nice if this macro supported `main` functions that don't return `()`. A main function can return anything that implements the `Termination` trait: https://doc.rust-lang.org/stable/std/process/trait.Termination.html.

Perhaps you could:

- run the main function code,
- pass the return value through a mutex,
- unpack the mutex, call `.report()` and return the exit code?

piodul (commented):
Please rebase on top of the "test macro" PR (I think we will merge the test macro first).

seastar/src/lib.rs
kfernandez31: Ah, I see.

seastar-macros/src/lib.rs
kfernandez31: Good idea, I didn't know of that trait. This shouldn't be too hard to implement.

seastar-macros/src/lib.rs
kfernandez31: `fn_token` is just the `fn` token. Agreed, the error should span the entire signature.

kfernandez31:
v2:
- support for all return types that derive `Termination`
- better spanning for errors
- other minor changes as requested above
Review ping @Lorak-mmk

seastar-macros/src/lib.rs
```diff
@@ -42,3 +42,70 @@ pub fn test(
...
+            let ret_holder_clone = ret_holder.clone();
+            let fut = async move {
+                let ret = #body;
+                ret_holder_clone.set(Some(ret));
+                Ok(())
+            };
+
+            let mut app = seastar::AppTemplate::default();
+            app.run_void(std::env::args(), fut);
+            ret_holder.take().unwrap()
```
Lorak-mmk, line 106: I think `Rc<Cell<Option<T>>>` is overkill here - oneshot channel would seem more appropriate.

Lorak-mmk (changes requested):

seastar-macros/src/lib.rs
Lorak-mmk: Now that I think of it, it's not necessarily true, current solution if pretty low overhead, channel may have some synchronization which would be more heavyweight. If there is some crate that has non-Send one-shot channel with comparable overhead to current solution, we could think about using it - if not, then let's stick with current solution.

kfernandez31: Here's a fast one I found: https://crates.io/crates/async-oneshot
And one of its benchmarks against `futures::oneshot::channel`: https://gist.github.com/jjl/4f7d7912194213577d39157b4f1192cc

kfernandez31: I benched `Rc<Cell<Option<T>` against `async_oneshot::oneshot::<T>` and here are the results:
### `Rc<Cell<Option<T>`
```
create_destroy          time:   [22.846 ns 22.879 ns 22.942 ns]        
send/success            time:   [39.096 ns 40.330 ns 41.581 ns]        
try_recv/success        time:   [38.105 ns 39.078 ns 40.054 ns]                             
```
### `async_oneshot::oneshot::<T>`
```
create_destroy          time:   [44.632 ns 44.688 ns 44.748 ns]                            
send/success            time:   [9.0997 ns 9.2451 ns 9.3948 ns]            
try_recv/success        time:   [41.189 ns 42.524 ns 43.838 ns]          
```
So from the perspective of the main thread, the first approach seems to be *slightly* faster.

Lorak-mmk: Ok, let's stick with `Rc<Cell<Option<T>>>` - it's inelegant, but I wouldn't want to sacrifice performance.

seastar/src/lib.rs
```diff
@@ -34,3 +34,25 @@ pub use preempt::*;
...
+///
+/// # Usage
+///
+/// ```rust
+/// #[seastar::main]
+/// async fn main() {
+///     println!("Hello, world!");
+/// }
+/// ```
+pub use seastar_macros::main;
```
piodul, line 58: Please add an example in the `examples` directory which runs a trivial program and uses the `main` macro.

piodul (commented):
Looks good to me after adding an example.

@Lorak-mmk please take a look, too.

Lorak-mmk (dismissed):

seastar/src/lib.rs
kfernandez31: Done.

Lorak-mmk (approved):

Issue #13, opened 2023-01-18
Ensure API safety when running code without having a runtime started


cvybhu:
We are aiming to make our API safe, which means that no matter what Rust code the user writes the program won't encounter any memory problems or undefined behavior.

When running seastar functions we should make sure that a runtime has been started. Without a runtime we might get a segfault, which isn't considered safe.

When a runtime isn't started we can just panic, that's what tokio does.

piodul:
Done in: https://github.com/zpp-2022-rust-seastar/seastar-rs/pull/27

PR #27, opened 2023-02-14
Ensuring api safety

kfernandez31:
Solves: #13 
Depends: #23

This PR introduces the `seastar::engine_is_ready` method for checking whether the current thread is in a Seastar runtime and the `seastar::assert_runtime_is_running!` and `seastar::assert_runtime_is_not_running!` macros that wrap over it.

We will want to utilize the macros in our APIs to ensure proper app behavior.
Example usage:

```diff
--- a/seastar/src/config_and_start_seastar.rs
+++ b/seastar/src/config_and_start_seastar.rs
@@ -217,6 +217,8 @@ impl AppTemplate {
     {
         let args: Vec<Arg> = args.into_iter().collect();
         let args: Vec<&str> = args.iter().map(|arg| arg.as_ref()).collect();
+        use crate as seastar;
+        assert_runtime_is_not_running!();
         run_void(self.app.pin_mut(), &args, VoidFuture::fallible_local(fut))
     }
```
```diff
--- a/seastar/src/submit_to.rs
+++ b/seastar/src/submit_to.rs
@@ -30,6 +30,8 @@ where
     Fut: Future<Output = Ret> + 'static,
     Ret: Send + 'static,
 {
+    use crate as seastar;
+    assert_runtime_is_running!();
     let (tx, rx) = futures::channel::oneshot::channel::<Ret>();
 
     let closure = move || {
```

[Are the macros hacky?](https://i.kym-cdn.com/photos/images/newsfeed/001/650/747/aaf.png). Well, for now there's no macro in Rust that returns the name of the current function, so some workarounds had to be done. A [PR](https://github.com/rust-lang/rfcs/pull/1719) and some [duplicates](https://github.com/rust-lang/rfcs/pull/2818) of it were open but none came to fruition.

### DIFF
--- a/seastar/src/api_safety.rs
+++ b/seastar/src/api_safety.rs
@@ -1,0 +1,62 @@
+#[cxx::bridge(namespace = "seastar")]
+mod ffi {
+    unsafe extern "C++" {
+        include!("seastar/core/reactor.hh");
+
+        /// Checks whether the current thread is within a Seastar runtime.
+        fn engine_is_ready() -> bool;
+    }
+}
+
+pub use ffi::engine_is_ready;
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::AppTemplate;
+    use std::thread;
+
+    #[test]
+    fn test_engine_is_ready_only_for_thread_in_runtime() {
+        assert!(!engine_is_ready());
+
+        thread::spawn(|| {
+            assert!(!engine_is_ready());
+            let _guard = crate::acquire_guard_for_seastar_test();
+            let mut app = AppTemplate::default();
+            let fut = async {
+                assert!(engine_is_ready());
+                Ok(())
+            };
+            let args = vec!["test"];
+            app.run_void(&args, fut);
+            // Currently, Seastar does not perform a cleanup of its thread locals,
+            // which unfortunately means the following assertion will pass.
+            // Should it fail, we must:
+            // 1) negate its condition
+            // 2) review other code in the project that also follows
+            //    this assumption and modify it accordingly.
+            assert!(engine_is_ready());
+        })
+        .join()
+        .unwrap();
+
+        assert!(!engine_is_ready());
+    }
+}
+
+/// Intended to be used in a runtime-dependent function.
+/// Panics if called outside of a Seastar runtime.
+pub fn assert_runtime_is_running() {
+    if !engine_is_ready() {
+        panic!("Attempting to call a runtime-dependent function outside of a Seastar runtime");
+    }
+}
+
+/// Mainly intended to be used in [`seastar::AppTemplate::run_void`] and  [`seastar::AppTemplate::run_int`].
+/// Panics if called within a Seastar runtime.
+pub fn assert_runtime_is_not_running() {
+    if engine_is_ready() {
+        panic!("Attempting to call a function inside of a Seastar runtime that must be used outside of it");
+    }
+}
--- a/seastar/build.rs
+++ b/seastar/build.rs
@@ -4,6 +4,7 @@ static CXX_BRIDGES: &[&str] = &[
     // Put all files that contain a cxx::bridge into this list
     "src/preempt.rs",
     "src/config_and_start_seastar.rs",
+    "src/api_safety.rs",
 ];
 
 static CXX_CPP_SOURCES: &[&str] = &["src/config_and_start_seastar.cc"];
--- a/seastar/src/lib.rs
+++ b/seastar/src/lib.rs
@@ -2,6 +2,7 @@
 //!
 //! Work in progress! Definitely not for use in production yet.
 
+mod api_safety;
 mod cxx_async_futures;
 mod cxx_async_local_future;
 
@@ -14,6 +15,7 @@ pub(crate) mod seastar_test_guard;
 #[cfg(test)]
 pub(crate) use seastar_test_guard::acquire_guard_for_seastar_test;
 
+pub use api_safety::*;
 pub use config_and_start_seastar::*;
 pub use preempt::*;
 


seastar/src/api_safety.rs
```diff
@@ -0,0 +1,82 @@
...
+            let mut app = AppTemplate::default();
+            let fut = async {
+                assert!(engine_is_ready());
+                Ok(())
+            };
+            app.run_void(std::env::args(), fut);
+            // Currently, Seastar does not perform a cleanup of its
+            // thread locals, which unfortunately means
+            // the following assertion  will pass.
+            assert!(engine_is_ready());
```
Lorak-mmk: If this problem is ever fixed is seastar, we will have to make some changes, like removing https://github.com/zpp-2022-rust-seastar/seastar-rs/blob/main/seastar/src/seastar_test_guard.rs
Please mention in the comment, that in a case this test fails, we need to look trough project's code and modify our assumptions accordingly - not just remove this line in the test,

seastar/src/api_safety.rs
```diff
@@ -0,0 +1,82 @@
...
+/// Intended to be used in a runtime-dependent function.
+/// Panics if called outside of a Seastar runtime.
+#[macro_export]
+macro_rules! assert_runtime_is_running {
+    () => {{
+        fn f() {}
+        fn type_name_of<T>(_: T) -> &'static str {
+            std::any::type_name::<T>()
+        }
+        let name = type_name_of(f);
```
Lorak-mmk: I don't understand this code, please explain what is going on here

seastar/src/api_safety.rs
```diff
@@ -0,0 +1,82 @@
...
+
+        if (!seastar::engine_is_ready()) {
+            panic!(
+                "Cannot use {} outside of a Seastar runtime",
+                &name[..name.len() - 3]
```
Lorak-mmk: Why `-3`?

Lorak-mmk (changes requested):
This will need a follw-up PR that uses those macros in all places that require them.

seastar/src/api_safety.rs
piodul: AFAIK most of the code here is a workaround to simulate something like `__FUNCTION__` in C. The type name for `f` function probably includes the name of the outer function and `f()` is later dropped with the `- 3` trick.

It's a bit hacky and I'm not sure it is guaranteed to work in the future. I'm not aware of any alternatives. Perhaps we could drop it? If somebody wants to check the source of the panic, they can set `RUST_BACKTRACE=1` or `RUST_BACKTRACE=full`.

piodul: There is also `std::backtrace` but its API is not really usable for this purpose yet: https://doc.rust-lang.org/std/backtrace/struct.Backtrace.html

seastar/src/api_safety.rs
```diff
@@ -0,0 +1,82 @@
...
+
+        thread::spawn(|| {
+            assert!(!engine_is_ready());
+            let _guard = crate::acquire_guard_for_seastar_test();
+            let mut app = AppTemplate::default();
+            let fut = async {
+                assert!(engine_is_ready());
+                Ok(())
+            };
+            app.run_void(std::env::args(), fut);
```
piodul: You shouldn't pass `std::env::args()` here. Seastar might not be able to understand the arguments that are passed to the test binary (e.g. if you try to run a single test).

seastar/src/api_safety.rs
Lorak-mmk: If this is a hacky workaround, the it should at least be properly explained in a comment, and it's not

seastar/src/api_safety.rs
kfernandez31: Agree, I'll change that.

seastar/src/api_safety.rs
kfernandez31: Sure, that's reasonable.

seastar/src/api_safety.rs
kfernandez31: Yes, it is precisely as @piodul said. How this works: we create a dummy function and get its type (and the type is qualified with the exact path to the function's declaration), we then cut `"::f` from the type to get the "path" to the parent-function, which in our case will be the one in which the assertion is called.

If you feel like this is unnecessary, and would rather see a more generic message like "invalid use of a runtime-dependent function outside a Seastar runtime", or no message at all (?), please do tell.

seastar/src/api_safety.rs
kfernandez31: see: the comment above.

seastar/src/api_safety.rs
kfernandez31: It's really unfortunate how a "`function!`" macro isn't in stable Rust, but `line!`, `column!`, `module_path!` and `file!` are...

Lorak-mmk: I think we don't need that - `RUST_BACKTRACE=1` is enough

kfernandez31:
Do we prefer functions over macros? I kinda wrote the assertions as macros by convention, but now there's really no point in that.

piodul:
> Do we prefer functions over macros? I kinda wrote the assertions as macros by convention, but now there's really no point in that.

Yes, If a function is sufficient then I think it is preferable to a macro.

kfernandez31:
v2: 
- Converted the macro into a function
- the panic messages are more generic now
Review ping @piodul

Lorak-mmk (approved):

piodul (approved):

piodul:
Note for the future: the "Fixes:" annotation - apart from letting us, humans correlate PR with the issue - is also a signal for GitHub to close the linked issue. It only works with "Fixes" as far as I know, so please use that word and don't use synonyms such as "Solves".

PR #29, opened 2023-02-15
Expose seastar::sleep and clocks to Rust

### LINKED ISSUE #28: Expose seastar::sleep to Rust
Make it possible to use `seastar::sleep` in Rust. The most basic specializations should be enough, as this function is a triple template.

patjed41:
Fixes: #28
Depends on: #24, #27

I made a lot of changes to this PR. The old description made no sense after applying these changes so I decided to write a new description.

This PR does 2 things now:
- Exposes specializations of clocks along with its duration and type_point types (`clocks` module).
- Uses 2 of 3 exposed duration types to implement Rust specializations of `seastar::sleep` (`sleep` module).

The unused part of the `clocks` module is here just for organizational purposes. It is used in the timer PR (#32).

# Clocks

The `clocks` module introduces:
- `SteadyClock` − equivalent of `std::chrono::steady_clock`,
- `LowresClock` − equivalent of `seastar::lowres_clock`,
- `ManualClock` − equivalent of `seastar::manual_clock`.

All of these clocks are unit-like structs with an associated function `now`. Additionally, `ManualClock` has a function `advance`.

## Durations and instants

Apart from the clocks, duration and type_point (called *instant* in Rust) types related to them are exposed. All duration types and all instant types have the same definition so we use `duration_impl` and `instant_impl` macros to generate all of them. These types are very similar to `std::time::{Duration, Instant}`, but use `i64` underneath. The duration types have a little bit less methods than `std::time::Duration`. For example, they lack methods using floats and saturating methods. I didn't implement them for now because it would be time-consuming and I don't think it would bring much value to our project. Implementing more methods can be done in the future if really needed.

So why are these types worth the effort of creating them? We can use the fact that `std::chrono::steady_clock::duration` is almost always defined as `std::chrono::nanoseconds` (even though it's unspecified in the C++ standard) and `std::chrono::nanoseconds::rep` almost always equals `int64_t` (it's also unspecified). We can make our API work perfectly, that is without any unwanted conversions, if these types are defined as written above. Moreover, we can use `std::chrono::duration_cast` if  `std::chrono::steady_clock::duration` is not defined as `std::chrono::nanoseconds`. The only hopeless case is when `std::chrono::steady_clock::duration::rep` does not equal `int64_t`. In such a case static assertions defined in `clocks.cc` fail the compilation.

# Sleeps

Creating duration types simplified `seastar::sleep` specializations a lot. We don't need `sleep_secs, sleep_millis, ...` now. One sleep function per clock is enough:
- `steady_sleep` using `SteadyClock`,
- `lowres_sleep` using `LowresClock`.


seastar/src/config_and_start_seastar.cc
```diff
@@ -54,4 +54,4 @@ int32_t run_int(app_template& app, int32_t argc, char** argv, IntFuture fut) {
 }
 
 } // namespace config_and_start_seastar
-} // namespace seastar
\ No newline at end of file
+} // namespace seastar
```
Lorak-mmk: This change is unrelated to this commit.
I don't have a problem with unrelated stylistic changed in a PR, but it should be in it's own commit

seastar/src/sleep.hh
```diff
@@ -0,0 +1,10 @@
+#pragma once
+
+#include "cxx_async_futures.hh"
+#include <seastar/core/sleep.hh>
+
+namespace seastar_ffi {
+namespace sleep {
+
+} // namespace sleep
+} // namespace seastar_ffi
```
Lorak-mmk: nit: no newline at end of file

seastar/src/lib.rs
```diff
@@ -7,6 +7,7 @@ mod cxx_async_local_future;
 
 mod config_and_start_seastar;
 mod preempt;
+#[allow(dead_code)]
```
Lorak-mmk: Why is it needed?

Lorak-mmk (changes requested):
Commit bdd4b5051acc4f1bd32f893b9e837df2c55501fd will have to be dropped after we merge #27

Please add some commit messages that explain shortly what is being changed and why.

seastar/src/config_and_start_seastar.cc
patjed41: Right, I moved this to a separate commit.

seastar/src/lib.rs
patjed41: It turns out it isn't needed. I had *unused function* warning for all sleep functions. I don't have them now. I'm not sure what changed but it works now.

patjed41:
Update:
- Made changes requested in a review.
- Added more verbose commit messages to commits where it seemed reasonable.

patjed41:
> Commit [bdd4b50](https://github.com/zpp-2022-rust-seastar/seastar-rs/commit/bdd4b5051acc4f1bd32f893b9e837df2c55501fd) will have to be dropped after we merge #27

That was my plan. Honestly, I didn't know that to do with this. My PR depends on 2 PRs: #24, #27. I can rebase my branch on one of them, but what about the second one? I can do another rebase or use cherry-pick but no matter what I do, creating a commit with changes from the second branch seems to be necessary (if it isn't I'd like to know how to achieve this). So I have 2 options:
- Create a commit that *picks* the changes from the seconds branch – in this specific case it is the *Pick API safety module* commit. Then create another commit that uses these changes to do what is needed – in this case it is the *Ensure API safety of sleep funtions* commit. If I do it like this, as pointed out, the first of these commits will have to be dropped after merging the second PR.
- Create a single commit that does both of the things described above at once. If I do it like this, I won't have to change the history of this PR because all of the changes picked from the second branch will already be here after merging that branch and will disappear from the list of changes introduced by this single commit.

So, my question is, what is the better approach in the case like this one?

piodul:
> * Create a commit that _picks_ the changes from the seconds branch – in this specific case it is the _Pick API safety module_ commit. Then create another commit that uses these changes to do what is needed – in this case it is the _Ensure API safety of sleep funtions_ commit. If I do it like this, as pointed out, the first of these commits will have to be dropped after merging the second PR.
> * Create a single commit that does both of the things described above at once. If I do it like this, I won't have to change the history of this PR because all of the changes picked from the second branch will already be here after merging that branch and will disappear from the list of changes introduced by this single commit.
> 
> So, my question is, what is the better approach in the case like this one?

Cherry picking is definitely better. After the cherry-picked commit is merged, you can then rebase on main. Sometimes, git is smart enough to notice this and will automatically skip that commit during interactive rebase.

Instead of picking and then fixing things after the picked commit, you can start your PR from the #24's branch, cherry pick #27 and then write your implementation with the fixes applied already.

seastar/src/sleep.rs
```diff
@@ -7,5 +7,38 @@ mod ffi {
...
+pub async fn sleep_micros(duration: i32) {
+    ffi::sleep_micros(duration).await.unwrap();
+}
+
+/// Returns a future which completes after a specified time in nanoseconds has elapsed.
+///
+/// Uses steady clock defined in Seastar.
+pub async fn sleep_nanos(duration: i64) {
+    ffi::sleep_nanos(duration).await.unwrap();
+}
```
piodul: Please prefix those with `steady_` - just like you did with `lowres_` functions, so that it is clear which timer those functions are using.

seastar/src/sleep.cc
```diff
@@ -3,5 +3,21 @@
...
+}
+
+VoidFuture sleep_millis(int32_t duration) {
+    co_await seastar::sleep(std::chrono::milliseconds(duration));
+}
+
+VoidFuture sleep_micros(int32_t duration) {
+    co_await seastar::sleep(std::chrono::microseconds(duration));
+}
+
```
piodul: I don't think that we need those. A single `sleep_nanos` should be sufficient. 2^64 nanoseconds is about 500-600 years, so `sleep_nanos` should be sufficient (for some time, at least).

piodul: Later, we can use `sleep_nanos` to build nicer, idiomatic abstractions - for example, introduce a `SteadyDuration` type which can be created from a number of seconds/milliseconds/etc. It would keep the duration as a number of nanoseconds so `sleep_nanos` could suffice.

seastar/src/sleep.cc
```diff
@@ -19,5 +19,13 @@ VoidFuture sleep_nanos(int64_t duration) {
     co_await seastar::sleep(std::chrono::nanoseconds(duration));
 }
 
+VoidFuture lowres_sleep_secs(int32_t duration) {
+    co_await seastar::sleep<seastar::lowres_clock>(std::chrono::seconds(duration));
+}
```
piodul: Same here - I think `lower_sleep_secs` is redundant.

seastar/src/sleep.rs
```diff
@@ -59,3 +59,57 @@ pub async fn lowres_sleep_secs(duration: i32) {
...
+        sleep_secs(duration).await;
+        let elapsed = now.elapsed().unwrap().as_secs() as i32;
+        assert_eq!(duration, elapsed);
+
+        // sleep_millis
+        let now = SystemTime::now();
+        let duration = 100;
+        sleep_millis(duration).await;
+        let elapsed = now.elapsed().unwrap().as_millis() as i32;
+        assert!(elapsed >= duration / 2 && elapsed < duration * 2);
```
piodul: I'm not sure about those. It's a bad practice in tests to expect for something to happen within a particular amout of time - the CI machines can sometimes be very slow and such assertions may lead to intermittent failures in CI.

IMO it should be fine to check that at least `duration` s/ms/ns passed, so maybe we could just stick to that?

seastar/src/sleep.rs
patjed41: Done.

seastar/src/sleep.cc
patjed41: After introducing duration types this issue does not exist anymore.

seastar/src/sleep.rs
patjed41: Right, I changed tests to check whether `elapsed >= duration`.

seastar/src/sleep.cc
patjed41: > Later, we can use `sleep_nanos` to build nicer, idiomatic abstractions - for example, introduce a `SteadyDuration` type which can be created from a number of seconds/milliseconds/etc. It would keep the duration as a number of nanoseconds so `sleep_nanos` could suffice.

I defined instant and duration types for all 3 clocks specializations.

There was a suggestion in a review of #32 that we could use `std::time::Duration`. It could be really easy to implement but I decided that creating our own duration types is worth the effort. I'll try to explain why:
- We would have to check if a given duration does not overflow `i64` in every function that uses duration. It is not a big problem currently since only sleeps and timer use durations but they might appear in other places (like `seastar::semaphore`) in the future if this project grows. Having to remember that checks are necessary wouldn't be nice.
- Something like *dont't use durations oveflowing `i64`* in documentation wouldn't look good. Readers would wonder why an overflowing type like this is even used.
- Different instant types using the same duration type would look very suspicious. In C++ the type_point types are dependant on the duration types.
- Overall, I think an API with our own duration types looks better. This is not a very signifacant difference but creating a type like that is not that hard and time-consuming so I don't see a reason why we wouldn't want it.

> I don't think that we need those. A single `sleep_nanos` should be sufficient. 2^64 nanoseconds is about 500-600 years, so `sleep_nanos` should be sufficient (for some time, at least).

After introducing duration types this issue does not exist anymore.

seastar/src/sleep.cc
patjed41: Defining duration types solved this.

patjed41:
Update:
- Made changes requested in a review.
- Introduced the `clocks` module (see a new description of this PR). These changes make part of #32 that exposes clocks outdated. I'll be working on fixing that PR from now.

seastar/src/clocks.rs
```diff
@@ -6,3 +8,135 @@ mod ffi {
...
 
 use ffi::*;
+
+// Macro used to generate the implemenetation of the instant types.
+// - `InstantName` - name of the instant type,
+// - `Instant` - must be equal to `InstantName`,
+// - `Duration` - name of the corresponding duration type,
+// - `Clock` - name of the corresponding clock.
+macro_rules! instant_impl {
+    ($InstantName:ident, $Instant:ty, $Duration:ty, $Clock:ty) => {
```
Lorak-mmk: Do we need both `InstantName` and `Instant`? Can't we use one of them in all the places?

seastar/src/clocks.rs
```diff
@@ -6,3 +8,135 @@ mod ffi {
...
+
+            /// Returns the amount of time elapsed from another instant to this one.
+            /// If `other` is later that `&self` the returned value is negative.
+            ///
+            /// # Panics
+            /// Panics if an overflow in the underyling data structure happens.
+            pub fn duration_since(&self, other: Self) -> $Duration {
+                match self.nanos.checked_sub(other.nanos) {
+                    Some(nanos) => <$Duration>::from_nanos(nanos),
+                    None => panic!(),
```
Lorak-mmk: All of the functions in this impl return options, but this one panics. Why the inconsistency?

seastar/src/clocks.rs
```diff
@@ -6,3 +8,135 @@ mod ffi {
...
+
+            /// Returns the amount of time elapsed from another instant to this one.
+            /// If `other` is later that `self` the returned value is negative.
+            ///
+            /// # Panics
+            /// Panics if an overflow in the underyling data structure happens.
+            fn sub(self, other: $Instant) -> $Duration {
+                self.duration_since(other)
+            }
+        }
```
Lorak-mmk: `AddAssing` and `SubAssign` have to panic on overflow, but `Add` and `Sub` don't. Considering that most of the functions in normal `impl` don't panic, maybe it's a better approach to make `Add` and `Sub` not panicking? Not sure what's the best approach here - it's a question of reliability vs ease of use. @piodul ?

seastar/src/clocks.rs
```diff
@@ -140,3 +140,244 @@ macro_rules! instant_impl {
...
+            /// # Panics
+            /// Panics if overflow happens.
+            fn neg(self) -> Self {
+                if self.nanos == i64::MAX {
+                    panic!();
+                }
+                Self { nanos: -self.nanos }
+            }
+        }
+    };
```
Lorak-mmk: Here it's the same consideration: should non-Assing version panic? @piodul

seastar/src/clocks.rs
```diff
@@ -140,3 +140,244 @@ macro_rules! instant_impl {
...
+        impl Neg for $Duration {
+            type Output = Self;
+
+            /// # Panics
+            /// Panics if overflow happens.
+            fn neg(self) -> Self {
+                if self.nanos == i64::MAX {
+                    panic!();
+                }
+                Self { nanos: -self.nanos }
```
Lorak-mmk: It doesn't look correct. Did you min `i64::MIN`? To be sure it works you could do 0.checked_sub(self.nanos)

seastar/src/clocks.rs
```diff
@@ -140,3 +140,244 @@ macro_rules! instant_impl {
...
+            }
+        }
+
+        impl Add<$Duration> for $Duration {
+            type Output = Self;
+
+            /// # Panics
+            /// Panics if result overflows.
+            fn add(self, rhs: Self) -> Self {
+                match self.nanos.checked_add(rhs.nanos) {
```
Lorak-mmk: Why not `self.checked_add`?
Then it could be just `self.checked_add(&rhs).unwrap()`

The same applies to all the other impls.

Lorak-mmk (changes requested):
Left some comments, overall looks pretty good.

It would be nice if there were some comments in the tests explaining what is the test testing etc - for example, line 534, `assert!(catch_unwind(|| i3.duration_since(i2)).is_err());`: it's not obvious on the first glance why this should panic, so a comment explaining that i3 is i64::MAX, i2 is less than zero, thus difference exceeds i64::MAX would be helpful. 
Such things may be obvious to the author right now, but it's hard to debug complicated tests later on when you can't understand them.

seastar/src/clocks.rs
patjed41: The problem was that with `$Instant:ty`
```rust
pub struct $Instant {
    pub(crate) nanos: i64,
}
```
didn't compile and I had to use `ident` instead of `ty` there.

I tried the other way − `ident` in all places − and it works. So I changed this macro to use `$Instant:ident, $Duration:ty, $Clock:ty`.

seastar/src/clocks.rs
patjed41: I based my instant types on `std::timer::Instant`. This type also has *checked* and *unchecked* versions of `duration_since`:
- [std::timer::Instant::duration_since](https://doc.rust-lang.org/std/time/struct.Instant.html#method.duration_since),
- [std::timer::Instant::checked_duration_since](https://doc.rust-lang.org/std/time/struct.Instant.html#method.checked_duration_since).

What's quite surprising is that `std::timer::Instant::duration_since` panics or saturates depending on the Rust version. I decided to consistently panic (instead of saturating) in all methods of my instant and duration types.

Implementing `duration_since` probably isn't necessary. I was simply implementing all methods that `std::timer::Instant` implements. So, in conclusion, my instant types are like `std::timer::Instant` with `std::chrono::type_point`'s features:
- they use a signed integer underneath,
- they implement `duration_since_epoch` (I called it like that instead of `time_since_epoch` because the rest of the methods use *duration*).

seastar/src/clocks.rs
patjed41: Yes, it was incorrect. Using `0.checked_sub(self.nanos)` is a really good idea. For some reason I had to write `0_i64.checked_sub(self.nanos)` to make it compile.

seastar/src/clocks.rs
patjed41: Good idea, simplified the code a lot.

seastar/src/clocks.rs
piodul: Let's stick to the conventions established in the Rust standard library. I can see that adding/subtracting instants/durations do not return options, so we should panic.

seastar/src/clocks.rs
```diff
@@ -6,3 +8,134 @@ mod ffi {
...
+            /// Returns `Some(t)` where `t` is equal to `self + duration` if `t` is
+            /// inside the bounds of the underlying data structure, None otherwise.
+            pub fn checked_add(&self, duration: $Duration) -> Option<Self> {
+                match self.nanos.checked_add(duration.nanos) {
+                    Some(nanos) => Some(Self { nanos }),
+                    None => None,
+                }
+            }
+
+            /// Returns `Some(t)` where `t` ii equal to `self - duration` if `t` is
```
piodul: Typo: `ii` -> `is`

seastar/src/clocks.rs
```diff
@@ -6,3 +8,134 @@ mod ffi {
 }
 
 use ffi::*;
+
+// Macro used to generate the implemenetation of the instant types.
```
piodul: Typo: `implemenetation` -> `implementation`

seastar/src/clocks.rs
```diff
@@ -6,3 +8,134 @@ mod ffi {
...
+        /// is is implemented as if it stores a value of
+        #[doc = concat!(" [`", stringify!($Duration), "`] ")]
+        /// indicating the time interval from the start of the
+        #[doc = concat!(" [`", stringify!($Clock), "`]'s epoch.")]
+        #[derive(Eq, PartialEq, Ord, PartialOrd, Clone, Copy, Hash, Default, Debug)]
+        pub struct $Instant {
+            pub(crate) nanos: i64,
+        }
+
+        impl $Instant {
```
piodul: I think that most of the Instant's methods could be made const, right?

seastar/src/sleep.cc
```diff
@@ -3,5 +3,13 @@
 namespace seastar_ffi {
 namespace sleep {
 
+VoidFuture steady_sleep(int64_t nanos) {
+    co_await seastar::sleep(std::chrono::nanoseconds(nanos));
+}
+
+VoidFuture lowres_sleep(int64_t nanos) {
+    co_await seastar::sleep<seastar::lowres_clock>(std::chrono::nanoseconds(nanos));
+}
```
piodul, line 12: Is there a reason why you didn't add a manual clock sleep?

seastar/src/clocks.rs
Lorak-mmk: I see you fixed that in duration impls, but not in instant impls, is there a reason for that?

Lorak-mmk: By the way, after applying this suggestion, all the arithmetic impls `impl Add, impl Mul, impl SubAssign etc` for both instants and durations will no longer need to know about fields of the structures - they will just call methods. 
I'm wondering if we can't create better API by:
1. Creating a trait for instant and a trait for duration - with the methods that are currently in the main impl block.
2. Write arithmetic impls once for those traits.
3. Macro will then just need to generate the structure itself and an implementation of this new trait.

I see some advantages of this approach:
- The biggest one: user will be able to write methods that are generic over instants / durations.
- Less source code after macro expansion (slightly faster compilation)
- Not sure about amount of generated code: possibly there is less of it, as rustc won't have to generate code for unused impl specializations?

seastar/src/clocks.rs
patjed41: I think all of them can be const and I changed all of them to const.

seastar/src/sleep.cc
patjed41: For some unknown reason I though `seastar::sleep` supports only steady and manual clocks. Of course, manual clock sleep can be added and I did that.

There is a slight problem with testing `manual_sleep`. I would like to spawn a task using `seastar::spawn` that advances `ManualClock` in a loop and then call and await `manual_sleep`. Unfortunately, I can't await `seastar::spawn` here and without awaiting `seastar::spawn` currently doesn't have any effect. I can write a complete test after someone fixes that. That's why there is only a very simple smoke test of `manual_sleep` now.

Edit: `seastar::spawn` was fixed and I implemented a complete test for `manual_sleep`.

seastar/src/clocks.rs
patjed41: > I see you fixed that in duration impls, but not in instant impls, is there a reason for that?

I forgot about that. It's fixed now.

After thinking about this new API for a moment I don't see a reason why this wouldn't work. Couldn't we also create a `Clock` trait that would implement only `now` function? Manual clock is a little different so maybe it's not the best idea.

I'm not sure if we want to implement this API and rewriting our current API seems to be a bit time-consuming so I've refrained from doing it for now. Maybe we will discuss this during tomorrow's meeting.

patjed41:
Update – generic `Duration`, `Instant` and `Sleep`:
- All clocks implement the `Clock` trait now. It contains only a `now` function.
- There is also a hidden `ClockImpl` trait that is a supertrait of `Clock`. All clock specific ffi functions should be put in this trait to avoid exposition outside the crate. There is only the `sleep` function for now. If we decide to stick to this solution, all ffi functions used by `Timer` will also go there.
- `ffi` module of the `timer` module was moved to the `clocks` module because it is used there.
- The code of `Duration`, `Instant` and `Sleep` is fully generic now.

I think current API has no significant disadvantages from the user's point of view. The code is fully generic, everything that should be public is public and everything that should be private is private. However, there is a small issue regarding the internal structure of the project. The `ClockImpl` trait forces part of the other modules' ffi (in this case its `sleep::ffi`) to be inside the `clocks` module which doesn't seem perfect. Keeping sleep's ffi in the `sleep` module and then `pub(crate)` exporting it would work here. However, problems would occur in the future if we did the same thing with timer's ffi. I can elaborate on that if we really want to avoid putting sleep's and timer's ffi in the `clocks` module.

Another issue with current changes is that putting
```rust
#[derive(Eq, PartialEq, Ord, PartialOrd, Hash, Clone, Copy, Default, Debug)]
```
before declaration of `Duration` and `Instant` doesn't work because of the `_phantom` field. I implemented all these traits manually. It could be shortened a little with the `derivative` crate but at the same time we would get another crate dependency and more complicated code.

Lorak-mmk (approved):

PR #32, opened 2023-02-20
Expose seastar::timer to Rust

### LINKED ISSUE #9: Expose seastar::timer to Rust
[`seastar::timer`](https://docs.seastar.io/master/classseastar_1_1timer.html) is a class that allows to asynchronously wait for a set amount of time. After this amount of time passes a callback is called.

There are two main challenges that come with implementing the timer.

1. Choosing an API on the Rust side.  `seastar::timer` API  speaks in time points, callbacks and arming. On the other hand other Rust runtimes prefer to expose an easy to use interface that can be `.awaited` on, for example [`tokio::time::sleep`](https://docs.rs/tokio/latest/tokio/time/fn.sleep.html). We need to decide how our time API will look like.

2. `seastar::timer` is parametrized using a `Clock` type.
    The seastar timer is designed in a way that allows to use it with different types of Clocks. Some clocks might be more efficient but lower resolution, other might be extremely precise but very slow to use.
    The Clock is passed as a template type parameter, so we might have `timer<lowres_clock>` , `timer<steady_clock_type>`, etc. 
    The Clock type also provides the types used to represent time points and durations.
    Template parameters are a pain with `cxx`, so we might have to work around them somehow.
    Creating a few different Timer types should be enough for most purposes. For example `struct LowResTimer`, `struct SteadyClockTimer` etc...

patjed41:
Fixes: #9
Depends on: #24, #27, #28 (only tests)

This PR exposes all 3 `seastar::timer`'s specializations to Rust: `SteadyClockTimer`, `LowresClockTimer`, `ManualClockTimer`.

## Clocks

To make an API of a specific timer usable, we also need to expose the `now` function of the clock used by this timer. Without this, timers' methods receiving a `time_point` value wouldn't be really useful. That's why this PR additionally exposes to Rust:
- `SteadyClock` − equivalent of `std::chrono::steady_clock`,
- `LowresClock` − equivalent of `seastar::lowres_clock`,
- `ManualClock` − equivalent of `seastar::manual_clock`.

All these clocks are unit-like structs with an associated function `now`. Moreover, `LowresClockTimer` has a function `update` and `ManualClockTimer` has a function `advance`. Exposing `advance` is necessary as calling this function is the only way to make a `ManualClock` advance. I don't know whether we should expose `update`. If not, I will remove it.

## Durations and time points

In my opinion, this is the most subtle and tricky part. We have to somehow expose `duration` and `type_point` types of each clock because timers' methods take values of those types. I simply used `i64`. For example, I defined `std::chrono::steady_clock::{duration, time_point}` like this:
```rust
pub type SteadyClockDuration = i64;
pub type SteadyClockTimePoint = i64;
```
Using `i64` is the only reasonable way of exposing these types I can think of. We can't pass these types directly across ffi because they are opaque. Even if we did that by using a `UniquePtr` for example, how are we going to let users create, access and perform arithmetical operations on them? Creating values of these types from `i64` seems to be the only option. Instead of defining aliases of `i64` we could make structs that create the *real* `duration` and `time_point` from `i64` but I don't think it would make much sense.

There is another issue with `i64`. Unfortunately, `std::chrono::steady_clock::duration` is unspecified in C++ standard. All of the most popular C++ libraries (`libc`, `libstdc++`, ...) define it as `std::chrono::nanoseconds` but it is still a non-standard. Moreover, `std::chrono::nanoseconds` is unspcecified too. More precisely, it is defined as `std::chrono::duration</*signed integer type of at least 64 bits*/, std::nano>`. So theoretically a conversion to a smaller type is possible if we use `i64`. I don't know if the issues described above are something we should be worried about.

## Timers

There are 3 timers:

- `SteadyClockTimer` − equivalent of `seastar::timer<std::chrono::steady_clock>`,
- `LowresClockTimer` − equivalent of `seastar::timer<seastar::lowres_clock>`,
- `ManualClockTimer` − equivalent of `seastar::timer<seastar::manual_clock>`.

### Callbacks

Callbacks are the only part of timer's API that is significantly changed after exposure to Rust. We allow closures implementing `FnMut()` trait. Note that `FnOnce()` trait is not sufficient because callbacks can be called multiple times. Passing callbacks to C++ is achieved by boxing them and converting to `*mut u8` − a substitue for `*mut c_void` that is not supported by `cxx`. Apart from the callback we pass to C++:
- caller − created by `get_caller` function, called from C++ to execute the callback.
- dropper - created by `get_dropper` function, called from C++ in an appropriate time to drop the callback.

We would like to drop the callback when a destructor of a function stored as timer's callback is called. So, instead of simply passing the callback to `timer.set_calback`, we pass a callable object (an instance of the `callback_object` class defined in `timer.cc`) that drops the callback by calling a dropper when its destructor is called. We have to make sure that a destructor of a temporary instance of `callback_object` created in `sct_set_callback` doesn't drop the callback. That's why we overwrite `callback_object`s move constructor and move operator.

### Arms and reams

The whole arm/rearm API is very similar to the C++ one. These are the changed I made:
- Methods using `time_point` are named with `until` infix. I needed to change name of some methods because we can't have 2 `arm` methods in Rust (one using `time_point` and one using `duration`).
- Timer has methods that take `std::optional<duration> period`. I could translate this to `Option<SomeClockDuration> period` but I decided to split these methods into non-periodic one and periodic one. I think API like is more pleasant to use. Of course, this can be changed.
- I added non-periodic rearm method that takes the duration type to each timer. From all 8 configurations of `{arm, rearm} x {periodic, non-periodic} x {takes duration, takes time_point}` this was the only one unimplemented. I just found it strange.

### Armed, cancel, get_timeout

The same as in C++.

## Tests

Every function exposed to Rust is tested. I think that the tests are strong enough but I can add more if needed. Additionally, I've written a doctest for each specialization of a timer.

## Scheduling group

Timer has a constructor and `set_callback` method that take an instance of `scheduling_group`. This is currently not exposed to Rust. I'm going to expose simplified `scheduling_group` to Rust and use it to implement the lacking part of the timer API in another PR.

## Structure of the timer module

I've put all the code in `timer.{rs, cc, hh}`. The problem is that the code is quite long and will become even longer after exposing `scheduling_group`. Should I keep it like this or split it somehow? Maybe putting each timer's specialization into a separate module and creating a common module (mainly for callers and droppers) is a good idea. I'm also not sure if clocks should be in the same place as timers.


seastar/src/timer.rs
```diff
@@ -252,6 +261,50 @@ impl SteadyClockTimer {
...
+    pub fn now() -> LowresClockTimePoint {
+        lowres_clock_now()
+    }
+
+    /// Updates `LowresClock`.
+    ///
+    /// Equivalent of `seastar::lowres_clock::update`.
+    pub fn update() {
+        lowres_clock_update()
+    }
```
piodul: I don't think we should expose this. AFAIK this function is only called by the runtime, as it is responsible for updating the lowres clock. Please remove it.

piodul:
Thanks for the detailed write-up!

> All these clocks are unit-like structs with an associated function `now`. Moreover, `LowresClockTimer` has a function `update` and `ManualClockTimer` has a function `advance`. Exposing `advance` is necessary as calling this function is the only way to make a `ManualClock` advance. I don't know whether we should expose `update`. If not, I will remove it.

I'm not sure we need `lowres_clock::update`. The runtime is responsible for updating the `lowres_clock`, and I don't see any uses of it in Scylla. I think it's better to remove it.

> 
> ## Durations and time points
> In my opinion, this is the most subtle and tricky part. We have to somehow expose `duration` and `type_point` types of each clock because timers' methods take values of those types. I simply used `i64`. For example, I defined `std::chrono::steady_clock::{duration, time_point}` like this:
> 
> ```rust
> pub type SteadyClockDuration = i64;
> pub type SteadyClockTimePoint = i64;
> ```
> 
> Using `i64` is the only reasonable way of exposing these types I can think of. We can't pass these types directly across ffi because they are opaque. Even if we did that by using a `UniquePtr` for example, how are we going to let users create, access and perform arithmetical operations on them? Creating values of these types from `i64` seems to be the only option. Instead of defining aliases of `i64` we could make structs that create the _real_ `duration` and `time_point` from `i64` but I don't think it would make much sense.

While it makes sense to use a `i64` number of milliseconds to represent a duration/time point, it's not ideal as far as type safety goes. Some operations on `i64` just don't make sense in the context of a duration, e.g. what does it mean for one duration to be multiplied by another?

I think that we should introduce dedicated types for representing time points:

```rust
// Or maybe call it SteadyClockInstant? it would align better with std::time::Instant
pub struct SteadyClockTimePoint {
    nanoseconds: i64,
}
```

You can implement arithmetic operators by implementing the right traits. Look at the [traits](https://doc.rust-lang.org/std/time/struct.Instant.html#trait-implementations) implemented by `std::time::Instant` - Add, AddAssign, Sub, SubAssign. We could implement similar things for our time points.

As far as duration goes, I think we could just use `std::time::Duration` as a representation. When passing through ffi, you will have to convert to and from `i64` nanoseconds (you can panic if Duration does not fit in i64).

> 
> There is another issue with `i64`. Unfortunately, `std::chrono::steady_clock::duration` is unspecified in C++ standard. All of the most popular C++ libraries (`libc`, `libstdc++`, ...) define it as `std::chrono::nanoseconds` but it is still a non-standard. Moreover, `std::chrono::nanoseconds` is unspcecified too. More precisely, it is defined as `std::chrono::duration</*signed integer type of at least 64 bits*/, std::nano>`. So theoretically a conversion to a smaller type is possible if we use `i64`. I don't know if the issues described above are something we should be worried about.

You can use [`std::chrono::duration_cast`](https://en.cppreference.com/w/cpp/chrono/duration/duration_cast) to make sure that the duration is converted to `std::chrono::nanoseconds`.

seastar/src/timer.cc
```diff
@@ -3,6 +3,58 @@
...
 namespace timer {
 
+// This class was created to drop callbacks passed to timer::set_calback at the right
+// time. We pass an instance of callback_object instead of a pure callback and the
+// callback is dropped when the callback_object's destrucotr is called.
+class callback_object {
+    uint8_t* _callback;
+    rust::Fn<void(uint8_t*)> _caller;
+    rust::Fn<void(uint8_t*)> _dropper;
+    bool _moved = false;
```
piodul: Nit: I'm not sure that `_moved` is the right name. For example, the default-constructed callback_object sets it to `true`, but it wasn't really moved from. How about `_valid` (+ flipping its value)?

seastar/src/timer.rs
```diff
@@ -12,11 +12,36 @@ mod ffi {
...
+
+fn dropper<Func: FnMut()>(raw_func: *mut u8) {
+    unsafe {
+        let _ = Box::from_raw(raw_func as *mut Func);
+    }
+}
+
+const fn get_dropper<Func: FnMut()>(_: &Func) -> fn(*mut u8) {
+    dropper::<Func>
+}
```
piodul: Nit: the `FnMut()` bounds here are unnecessary.

seastar/src/timer.rs
```diff
@@ -110,4 +127,127 @@ impl SteadyClockTimer {
...
+    /// Sets the timer expiration time with automatic rearming.
+    ///
+    /// It is illegal to arm a timer that has already been armed (and not disarmed by
+    /// expiration or [`SteadyClockTimer::cancel`]). In the current implementation,
+    /// this will result in an assertion failure. See
+    /// [`SteadyClockTimer::rearm_until_periodic`].
+    ///
+    /// # Arguments
+    /// * `until` - The time when the timer expires.
+    /// * `period` - Utomatic rearm duration.
```
piodul: Nit: `Utomatic` -> `Automatic`

seastar/src/timer.rs
```diff
@@ -110,4 +127,127 @@ impl SteadyClockTimer {
...
+    pub fn arm_until_periodic(&mut self, until: SteadyClockTimePoint, period: SteadyClockDuration) {
```
piodul: The `_until` part in `arm_until_periodic` suggests to me that the timer will be fired periodically _until_ some point in time. How about `arm_at_periodic`? And then changing the other functions to use `at` instead of `until`?

cc: @cvybhu @Lorak-mmk what do you think?

seastar/src/timer.rs
```diff
@@ -305,6 +333,168 @@ impl LowresClock {
     }
 }
 
+/// Specialization of Seastar's timer that uses [`LowresClock`] as a clock.
```
piodul: We have lots of code duplicated for all three types of clocks. I strongly suggest looking into declarative macros in order to generate slightly different implementations for all clock types.

I see this pattern sometimes when looking at the stdlib code. An example:

Macro definition: https://doc.rust-lang.org/src/core/num/int_macros.rs.html#1
Use: https://doc.rust-lang.org/src/core/num/mod.rs.html#241

seastar/src/timer.rs
patjed41: Removed.

seastar/src/timer.cc
patjed41: Yes, `_valid` is much better.

seastar/src/timer.rs
patjed41: Good point, I fixed it.

seastar/src/timer.rs
patjed41: I changed all `until`s to `at` for now.

seastar/src/timer.rs
patjed41: I made 2 macros:
- `timer_impl` that generates implementation of the timer specializations,
- `test_timer` that generates tests for the timer specializations.

The whole ffi part of the timer module still doesn't use macros. I don't think that using macros inside the `#[cxx::bridge]` module is possible. I didn't use macros in `timer.{hh, cc}` too. I'm not sure if using C++ macros there is a good idea. My guess is that it wouldn't look really good. For example, we would have to concatenate parts of the functions' names.

seastar/src/timer.rs
patjed41: Fixed.

patjed41:
Update:
- I made changes requested in a review.
- Using `timer_impl` and `test_timer` macros changed the code and this PR's commit history a lot. There are new commits and the number of all commits is much smaller.
- I moved the whole clock/duration/time_point thing to #29. Some duration types were needed there and the rest of the code was moved there for organisational purposes (I didn't want to split it into 2 PRs).
- I updated the description of this PR a little.
- To follow conventions used in Rust, I changed the name of the only field in `SomeClockTimer` structs from `timer` to `inner`.
- The C++ side of the ffi must convert duration from `std::chrono::nanoseconds` to duration types used by the clocks (and the other way around). This problem might appear in the future in any ffi that uses durations and instants. That's why I decided to solve this generally by creating a small conversion API in `clocks.hh`. I think it's the best place for the code like this as this is the most *general* place for the code related to the clocks.
- `set_callback` enforces `'static` lifetime on `Func` now.
- I implemented methods using duration with methods using instants (*at* methods) by using the `now` function and the `+` operator. It reduced a number of arm/rearm functions exposed through ffi.

seastar/src/timer.rs
```diff
@@ -7,3 +7,21 @@ mod ffi {
...
+
+fn dropper<Func>(raw_func: *mut u8) {
+    unsafe {
+        let _ = Box::from_raw(raw_func as *mut Func);
+    }
+}
+
+const fn get_dropper<Func>(_: &Func) -> fn(*mut u8) {
+    dropper::<Func>
+}
```
piodul: These should be put in the `ffi_utils` module, after this PR is merged: https://github.com/zpp-2022-rust-seastar/seastar-rs/pull/35

seastar/src/timer.rs
```diff
@@ -106,3 +107,160 @@ fn dropper<Func>(raw_func: *mut u8) {
...
+                /// In the current implementation, this will result in an assertion failure. See
+                #[doc = concat!(" [`", stringify!($Timer), "::rearm_periodic`].")]
+                ///
+                /// # Arguments
+                /// * `delta` - The time when the timer expires, relative to now.
+                pub fn arm_periodic(&mut self, delta: $Duration) {
+                    self.arm_at_periodic(<$Clock>::now() + delta, delta);
+                }
+
+                /// Sets the timer expiration time relatively to now. If the timer was already then,
```
piodul: `already then` -> `already armed`?

seastar/src/timer.rs
```diff
@@ -106,3 +107,160 @@ fn dropper<Func>(raw_func: *mut u8) {
...
+                ///
+                /// Returns `true` if the timer was armed before the call.
+                pub fn cancel(&mut self) -> bool {
+                    [<$ffi_pref _cancel>](self.inner.pin_mut())
+                }
+
+                /// Gets the expiration time of an armed timer.
+                pub fn get_timeout(&self) -> $Instant {
+                    $Instant { nanos: [<$ffi_pref _get_timeout>](&self.inner) }
+                }
```
piodul: From the original docstring:
```rust
    /// \return the time at which the timer is scheduled to expire (undefined if the
    ///       timer is not armed).
```
While it's not undefined behavior, returning an undefined value is not a good API IMO. Let's slighly improve the semantics of the original function and return an Option here - if the timer is not armed then it will return None.

seastar/src/timer.rs
patjed41: Done. `caller` is called `fn_mut_void_caller` now. I don't think it's that good but it can't be called `fn_mut_caller` because it does not return any type like `fn_once_caller`.

patjed41:
Update − fully generic `Timer`:
- Analogous to `sleep` (#29), all clock specific ffi functions are implemented inside the `ClockImpl` trait and `ffi` module is moved from the `timer` module to the `clocks` module.
- `Timer`'s field called `inner` has different type for different timer specializations. However, `ClockImpl` forces us to use the same type for all specializations. Enum `CppTimer` solves this problem. Another approach could be to cast `inner` to `*mut u8` or something but I think enum is better here.
- The only significant issue I see here is that the timer's ffi is inside the `clocks` module (the same problem is with `sleep`).

seastar/src/clocks.rs
```diff
@@ -550,13 +551,139 @@ impl<ClockType> Sub for Instant<ClockType> {
     }
 }
 
+#[doc(hidden)]
+pub enum CppTimer {
+    Steady(UniquePtr<steady_clock_timer>),
+    Lowres(UniquePtr<lowres_clock_timer>),
+    Manual(UniquePtr<manual_clock_timer>),
+}
```
piodul: Hmm, why do we have a single `CppTimer`? I thought we would have a separate Rust type for each seastar clock. You could then expose this type via associated type in `ClockImpl` so that the generic `Timer` that you introduce in the next commit will have access to it.

patjed41: I fixed it.

piodul (approved):

PR #35, opened 2023-02-24
Submit to

### LINKED ISSUE #15: Implement seastar::submit_to
`seastar::submit_to` is a function that allows to run code on a specific shard.

We would like to be able to do that from Rust.
It would be similar to #5, but the function to run will most likely need to be `Send` as it will be sent to a different core.

grzenow4:
Fixes: #15 
Depends on: #2 

This PR introduces a way to run code on a specific shard by method:
```rust
pub async fn submit_to<Func, Fut, Ret>(shard_id: u32, func: Func) -> Ret
where
    Func: FnOnce() -> Fut + Send + 'static,
    Ret: Send + 'static,
```
For example:
```rust
let _ = submit_to(0, || async { 42 }).await;
```

### DIFF
--- a/seastar/src/submit_to.hh
+++ b/seastar/src/submit_to.hh
@@ -1,0 +1,17 @@
+#pragma once
+
+#include "cxx-async/include/rust/cxx_async_seastar.h"
+#include "rust/cxx.h"
+#include "cxx_async_futures.hh"
+
+namespace seastar_ffi {
+
+namespace submit_to {
+
+using fn_once_caller_t = VoidFuture (*)(const uint8_t*);
+
+VoidFuture submit_to(const uint32_t shard_id, uint8_t* closure, rust::Fn<VoidFuture(uint8_t*)> caller);
+
+} // submit_to
+
+} // seastar_ffi
--- a/seastar/src/submit_to.rs
+++ b/seastar/src/submit_to.rs
@@ -1,0 +1,95 @@
+use crate as seastar;
+use crate::ffi_utils::{get_dropper, get_fn_once_caller};
+use ffi::*;
+use std::future::Future;
+
+use crate::cxx_async_local_future::IntoCxxAsyncLocalFuture;
+
+#[cxx::bridge]
+mod ffi {
+    #[namespace = "seastar_ffi"]
+    unsafe extern "C++" {
+        type VoidFuture = crate::cxx_async_futures::VoidFuture;
+    }
+
+    #[namespace = "seastar_ffi::submit_to"]
+    unsafe extern "C++" {
+        include!("seastar/src/submit_to.hh");
+
+        unsafe fn submit_to(
+            shard_id: u32,
+            closure: *mut u8,
+            caller: unsafe fn(*mut u8) -> VoidFuture,
+        ) -> VoidFuture;
+    }
+}
+
+/// Runs a function `func` on a `shard_id` shard.
+///
+/// # Example
+///
+/// ```rust
+/// #[seastar::test]
+/// async fn submit_to_example() {
+///     let ret = submit_to(0, || async { 42 }).await;
+///     assert!(matches!(ret, 42));
+/// }
+pub async fn submit_to<Func, Fut, Ret>(shard_id: u32, func: Func) -> Ret
+where
+    Func: FnOnce() -> Fut + Send + 'static,
+    Fut: Future<Output = Ret> + 'static,
+    Ret: Send + 'static,
+{
+    let (tx, rx) = futures::channel::oneshot::channel::<Ret>();
+
+    let closure = move || {
+        VoidFuture::infallible_local(async {
+            tx.send(func().await).ok();
+        })
+    };
+
+    let closure_caller = get_fn_once_caller(&closure);
+    let dropper = get_dropper(&closure);
+    let boxed_closure = Box::into_raw(Box::new(closure)) as *mut u8;
+
+    unsafe {
+        match ffi::submit_to(shard_id, boxed_closure, closure_caller).await {
+            Ok(_) => rx.await.unwrap(),
+            Err(_) => {
+                dropper(boxed_closure);
+                panic!()
+            }
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[seastar::test]
+    async fn test_submit_to() {
+        let ret = submit_to(0, || async { 42 }).await;
+        assert!(matches!(ret, 42));
+    }
+
+    #[seastar::test]
+    async fn test_submit_to_nested() {
+        let ret = submit_to(0, || async { submit_to(0, || async { 42 }).await }).await;
+        assert!(matches!(ret, 42));
+    }
+
+    #[seastar::test]
+    async fn test_submit_to_two_shards() {
+        let ret = submit_to(0, || async { 17 }).await;
+        assert!(matches!(ret, 17));
+        let ret = submit_to(1, || async { 25 }).await;
+        assert!(matches!(ret, 25));
+    }
+
+    #[seastar::test]
+    async fn test_submit_to_two_shards_nested() {
+        let ret = submit_to(1, || async { submit_to(0, || async { 42 }).await }).await;
+        assert!(matches!(ret, 42));
+    }
+}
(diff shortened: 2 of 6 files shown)


seastar/src/submit_to.rs
```diff
@@ -0,0 +1,110 @@
...
+    unsafe extern "C++" {
+        type VoidFuture = crate::cxx_async_futures::VoidFuture;
+    }
+
+    #[namespace = "seastar_ffi::submit_to"]
+    unsafe extern "C++" {
+        include!("seastar/src/submit_to.hh");
+
+        unsafe fn submit_to(
+            _shard_id: u32,
```
kfernandez31: I believe this wasn't intended to be prefixed with an underscore.

seastar/src/submit_to.rs
```diff
@@ -0,0 +1,110 @@
...
+/// async fn submit_to_example() {
+///     let ret = submit_to(0, || future::ready(42)).await;
+///     assert!(ret.is_ok());
+///     assert_eq!(ret.unwrap(), 42);
+/// }
+pub async fn submit_to<Func, Fut, Ret>(shard_id: u32, func: Func) -> cxx_async::CxxAsyncResult<Ret>
+where
+    Func: FnOnce() -> Fut + Send + 'static,
+    Fut: Future<Output = Ret> + 'static,
+    Ret: Send + std::fmt::Debug + 'static,
```
kfernandez31: Why the requirement for `std::fmt::Debug`?

seastar/src/submit_to.rs
```diff
@@ -0,0 +1,110 @@
...
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use futures::future;
+
+    #[seastar::test]
+    async fn test_submit_to() {
+        let ret = submit_to(0, || future::ready(42)).await;
+        assert!(ret.is_ok());
```
kfernandez31: These two assertions can be combined into one: `assert!(matches!(res, Ok(0))`

seastar/src/submit_to.rs
```diff
@@ -0,0 +1,110 @@
...
+            Err(e) => {
+                dropper(boxed_closure);
+                Err(e)
+            }
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
kfernandez31, line 67: I think it'd be good to have a test of nested `submit_to`s on top of the existing one.

seastar/src/submit_to.rs
```diff
@@ -0,0 +1,110 @@
...
+
+/// Convert the pointer to Box<Func> and then call it, consuming it in the process.
+fn fn_once_caller<Func>(raw_func: *mut u8) -> VoidFuture
+where
+    Func: FnOnce() -> VoidFuture,
+{
+    unsafe { Box::from_raw(raw_func as *mut Func)() }
+}
+
+/// A helper function. We need to be able to name the type of a named closure.
```
kfernandez31: Nit: if we're commenting this, we could also comment on `fn_once_dropper`, or mention it here, that the reason for it is identical.

kfernandez31 (changes requested):

seastar/src/submit_to.rs
grzenow4: It is required by `std::result::Result::<T, E>::unwrap`.

seastar/src/submit_to.hh
```diff
@@ -0,0 +1,17 @@
...
+
+#include "cxx-async/include/rust/cxx_async_seastar.h"
+#include "rust/cxx.h"
+#include "cxx_async_futures.hh"
+
+namespace seastar_ffi {
+
+namespace submit_to {
+
+typedef VoidFuture (*fn_once_caller_t)(const uint8_t*);
```
piodul: Nit: please use `using` instead of `typedef`, it has much more sensible syntax:

```c++
using fn_once_caller_t = VoidFuture (*)(const uint8_t*);
```

seastar/src/submit_to.rs
```diff
@@ -0,0 +1,122 @@
...
+    async fn test_submit_to_two_shards() {
+        let ret = submit_to(0, || async { 17 }).await;
+        assert!(matches!(ret, Ok(17)));
+        let ret = submit_to(1, || async { 25 }).await;
+        assert!(matches!(ret, Ok(25)));
+    }
+
+    #[seastar::test]
+    async fn test_submit_to_two_shards_nested() {
+        let ret = submit_to(0, || async { submit_to(1, || async { 42 }).await.unwrap() }).await;
```
piodul: Nit: it would be slightly more interesting to submit the outer task to shard 1 and the inner to 0 - so that both submits need to cross the shard boundary.

seastar/src/submit_to.rs
```diff
@@ -0,0 +1,122 @@
...
+            closure: *mut u8,
+            caller: unsafe fn(*mut u8) -> VoidFuture,
+        ) -> VoidFuture;
+    }
+}
+
+/// Convert the pointer to Box<Func> and then call it, consuming it in the process.
+fn fn_once_caller<Func>(raw_func: *mut u8) -> VoidFuture
+where
+    Func: FnOnce() -> VoidFuture,
```
piodul: Nit: it could probably be made more generic:

```rust
fn fn_once_caller<Func, Ret>(raw_func: &mut u8) -> VoidFuture
where
    Func: FnOnce() -> Ret
{
    // ...
}
```

You can do something similar to `get_fn_once_caller`.

seastar/src/submit_to.rs
```diff
@@ -0,0 +1,122 @@
...
+{
+    fn_once_caller::<Func>
+}
+
+/// Free a pointer. We need to call this function in the case of an error.
+fn fn_once_dropper<Func>(raw_func: *mut u8) {
+    unsafe {
+        let _ = Box::from_raw(raw_func as *mut Func);
+    }
+}
```
piodul: Nit: the name is too specific IMO. It's suitable of dropping anything in a box, really.

seastar/src/submit_to.rs
```diff
@@ -0,0 +1,122 @@
...
+
+const fn get_fn_once_dropper<Func>(_: &Func) -> fn(*mut u8) {
+    fn_once_dropper::<Func>
+}
```
piodul: I anticipate that those functions might be useful in other contexts, e.g. there is some overlap with what the timers are doing. Please move them to a new (internal) module `ffi_utils`.

seastar/src/submit_to.rs
```diff
@@ -0,0 +1,122 @@
...
+/// Convert the pointer to Box<Func> and then call it, consuming it in the process.
+fn fn_once_caller<Func>(raw_func: *mut u8) -> VoidFuture
+where
+    Func: FnOnce() -> VoidFuture,
+{
+    unsafe { Box::from_raw(raw_func as *mut Func)() }
+}
+
+/// A helper function. We need to be able to name the type of a named closure.
+/// There is no `typeof` in Rust, so this is achieved by using a dummy paramter.
```
piodul: Nit: typo: `paramter` -> `parameter`

piodul:
(needs rebase, by the way)

grzenow4:
Done, it is ready to merge now.

seastar/src/submit_to.rs
grzenow4: Is `fn_dropper` and `get_fn_dropper` fine or shall I get rid of `fn` infix?

piodul:
Please rebase, again - there was an issue with the CI which has been fixed (https://github.com/zpp-2022-rust-seastar/seastar-rs/pull/36)

grzenow4:
Done.

seastar/src/submit_to.rs
piodul: Yes, please. The `Func` parameter name should also be changed to something more generic, e.g. `T`.

The comment

```rust
/// Free a pointer. We need to call this function in the case of an error.
```

is also too specific. As this function is now in a separate module and can be reused by others, the details of how it is used in `submit_to` are irrelevant (I mean the second sentence here).

grzenow4: All done.

piodul (approved):

PR #36, opened 2023-03-03
CI: run `apt update` before installing anything

piodul:
It is possible for the test runner's apt index to become stale, which can lead to apt trying to download files from URLs that don't exist. Fix that by always running `apt update` before continuing with installation of packages necessary for compilation.


Lorak-mmk (approved):

cvybhu (commented):
:+1:

PR #38, opened 2023-03-07
submit_to works without await

grzenow4:
This PR changes `submit_to` function signature to
```rust
pub fn submit_to<Func, Fut, Ret>(shard_id: u32, func: Func) -> impl Future<Output = Ret>
where
    Func: FnOnce() -> Fut + Send + 'static,
    Ret: Send + 'static,
```
Now you don't have to call the `await` to start a function execution. Simple test for this change is added:
```rust
#[seastar::test]
async fn test_submit_to_no_await() {
    let (tx, rx) = futures::channel::oneshot::channel::<i32>();
    let _ = submit_to(0, || async {
        tx.send(42).ok();
    });
    assert!(matches!(rx.await.unwrap(), 42));
}
```
Also small correction: `submit_to` checks if runtime is running.

### DIFF
--- a/seastar/src/submit_to.rs
+++ b/seastar/src/submit_to.rs
@@ -34,12 +34,15 @@ mod ffi {
 ///     let ret = submit_to(0, || async { 42 }).await;
 ///     assert!(matches!(ret, 42));
 /// }
-pub async fn submit_to<Func, Fut, Ret>(shard_id: u32, func: Func) -> Ret
+/// ```
+pub fn submit_to<Func, Fut, Ret>(shard_id: u32, func: Func) -> impl Future<Output = Ret>
 where
     Func: FnOnce() -> Fut + Send + 'static,
     Fut: Future<Output = Ret> + 'static,
     Ret: Send + 'static,
 {
+    crate::assert_runtime_is_running();
+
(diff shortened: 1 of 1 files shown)


Lorak-mmk (dismissed):

seastar/src/submit_to.rs
```diff
@@ -34,12 +34,15 @@ mod ffi {
...
 /// }
-pub async fn submit_to<Func, Fut, Ret>(shard_id: u32, func: Func) -> Ret
+/// ```
+pub fn submit_to<Func, Fut, Ret>(shard_id: u32, func: Func) -> impl Future<Output = Ret>
 where
     Func: FnOnce() -> Fut + Send + 'static,
     Fut: Future<Output = Ret> + 'static,
     Ret: Send + 'static,
 {
+    seastar::assert_runtime_is_running();
```
piodul: I'm not sure why, but the build fails on this line. Perhaps `crate::assert_runtime_is_running()` would work instead?

grzenow4: Indeed, it solves the issue.

piodul (approved):

PR #40, opened 2023-03-16
Expose `seastar::distributed`

### LINKED ISSUE #21: Make it possible to use seastar::distributed in Rust
`seastar::distributed` allows to run a service across all seastar shards.

It would be useful to be able to do this from Rust.

The problem is that `seastar::distributed` has a template argument which specifies the service to run.
We might have to design a different Rust API that is safe and easy to use.

kfernandez31:
Resolves: #21 

This PR exposes the [`seastar::distributed<S>`](https://github.com/scylladb/seastar/blob/master/include/seastar/core/distributed.hh) API, an alias for `seastar::sharded<S>`, as `Distributed<S>`.

In Seastar, the type over which `distributed` is generic must expose a `stop`, which is expressed by the following snippet in `sharded.hh`:
```cpp
// Helper check if Service::stop exists

struct sharded_has_stop {
    // If a member names "stop" exists, try to call it, even if it doesn't
    // have the correct signature. This is so that we don't ignore a function
    // named stop() just because the signature is incorrect, and instead
    // force the user to resolve the ambiguity.
    template <typename Service>
    constexpr static auto check(int) -> std::enable_if_t<(sizeof(&Service::stop) >= 0), bool> {
        return true;
    }

    // Fallback in case Service::stop doesn't exist.
    template<typename>
    static constexpr auto check(...) -> bool {
        return false;
    }
};
```
This was presumably written before the introduction of concepts in C++/Seastar, but that doesn't really matter. To enforce this rule in Rust, we simply use a trait:
```rs
pub trait Service {
    fn stop(&self) -> Box<dyn Future<Output = ()>> {
        Box::new(async {})
    }
}
```

## `Distributed`'s methods 
While `Distributed` exposes the `start_single`, `start` and `stop` methods, which are equivalent to their C++ counterparts, other methods differ slightly. That is the case for:
- `distributed::map`,
- `distributed::invoke_on_others`,
- `distributed::invoke_on`.

While `distributed::map_reduce` and `distributed::map_reduce0` are not exposed to avoid redundancy - we can just expose some mapping functions, and the user of the API can reduce its results by hand with the addition of a single method call (either `reduce` or `fold` from the `Iterator` trait).

But back to the aforementioned list - what is actually exposed is:
- `Distributed::map_all` - *almost* equivalent to `distributed::map`,
- `Distributed::map_others` - *almost* equivalent to `distributed::invoke_on_others` but can handle futures of a return type other than void/unit,
- `Distributed::map_single` - equivalent to `distributed::invoke_on` but, like above, handles all sorts of types.

The reason for diverging from the original methods was to provide a more powerful API. Because the original methods are not transferred directly, the new functions are implemented using `submit_to` (if I recall correctly, the C++ methods make use of it as well).  

The reason for writing "*almost*" in the above list is that due to each call of `submit_to` yielding a future, `map_all` and `map_others` return `Vec<impl Future>`. This change too leads to more flexibility, as the API user can decide for themselves whether they want to squash the futures together using something like `futures::future::join_all` or wait for merely a subset of them.

## A small dependency

The total number of shards, as well as the current one, must be known in order to delegate work to a subset of shards. Because of this, I added a new module in `smp.rs` that exposes functions for precisely that:
- `this_shard_id` - returns the id of the current shard,
- `get_count` - returns the total number of shards.

This may come in handy in some future features as well.

### DIFF
--- a/seastar/src/distributed.cc
+++ b/seastar/src/distributed.cc
@@ -1,0 +1,66 @@
+#include "distributed.hh"
+#include <seastar/util/defer.hh>
+
+namespace seastar_ffi {
+namespace distributed {
+
+rust_service::rust_service(
+    const uint8_t* raw_service_maker,
+    rust::Fn<uint8_t*(const uint8_t*)> raw_service_maker_caller,
+    rust::Fn<VoidFuture(uint8_t*)> stop_caller,
+    rust::Fn<void(uint8_t*)> dropper)
+: _stop_caller(stop_caller), _dropper(dropper)
+{
+    _inner = raw_service_maker_caller(raw_service_maker);
+}
+
+seastar::future<> rust_service::stop() {
+    co_await _stop_caller(_inner);
+}
+
+rust_service::~rust_service() {
+    _dropper(_inner);
+}
+
+std::shared_ptr<distributed> new_distributed() {
+    return std::make_shared<distributed>();
+}
+
+const uint8_t* local(const distributed& distr) {
+    return distr.local()._inner;
+}
+
+VoidFuture start(
+    const distributed& distr,
+    const uint8_t* raw_service_maker,
+    rust::Fn<uint8_t*(const uint8_t*)> raw_service_maker_caller,
+    rust::Fn<void(const uint8_t*)> raw_service_maker_dropper,
+    rust::Fn<VoidFuture(uint8_t*)> stop_caller,
+    rust::Fn<void(uint8_t*)> dropper
+) {
+    auto cleanup_maker = seastar::defer([&] {
+        raw_service_maker_dropper(raw_service_maker);
+    });
+    co_await const_cast<distributed&>(distr).start(raw_service_maker, raw_service_maker_caller, stop_caller, dropper);
+}
+
+VoidFuture start_single(
+    const distributed& distr,
+    const uint8_t* raw_service_maker,
+    rust::Fn<uint8_t*(const uint8_t*)> raw_service_maker_caller,
+    rust::Fn<void(const uint8_t*)> raw_service_maker_dropper,
+    rust::Fn<VoidFuture(uint8_t*)> stop_caller,
+    rust::Fn<void(uint8_t*)> dropper
+) {
+    auto cleanup_maker = seastar::defer([&] {
+        raw_service_maker_dropper(raw_service_maker);
+    });
+    co_await const_cast<distributed&>(distr).start_single(raw_service_maker, raw_service_maker_caller, stop_caller, dropper);
+}
+
+VoidFuture stop(const distributed& distr) {
+    co_await const_cast<distributed&>(distr).stop();
+}
+
+} // namespace distributed
+} // namespace seastar_ffi
--- a/seastar/src/distributed.rs
+++ b/seastar/src/distributed.rs
@@ -1,0 +1,517 @@
+use crate::{
+    cxx_async_local_future::IntoCxxAsyncLocalFuture,
+    ffi_utils::{get_dropper_const, get_dropper_noarg, get_fn_caller},
+    get_count,
+    submit_to::submit_to,
+    this_shard_id,
+};
+use core::marker::PhantomData;
+use cxx::SharedPtr;
+use std::future::Future;
+use std::pin::Pin;
+
+#[cxx::bridge(namespace = "seastar_ffi::distributed")]
+mod ffi {
+    unsafe extern "C++" {
+        include!("seastar/src/distributed.hh");
+
+        type distributed;
+
+        #[namespace = "seastar_ffi"]
+        type VoidFuture = crate::cxx_async_futures::VoidFuture;
+
+        fn new_distributed() -> SharedPtr<distributed>;
+
+        fn local(distr: &distributed) -> *const u8;
+
+        unsafe fn start(
+            distr: &distributed,
+            raw_service_maker: *const u8,
+            raw_service_maker_caller: unsafe fn(*const u8) -> *mut u8,
+            raw_service_maker_droppper: unsafe fn(*const u8),
+            stop_caller: unsafe fn(*mut u8) -> VoidFuture,
+            dropper: unsafe fn(*mut u8),
+        ) -> VoidFuture;
+
+        unsafe fn start_single(
+            distr: &distributed,
+            raw_service_maker: *const u8,
+            raw_service_maker_caller: unsafe fn(*const u8) -> *mut u8,
+            raw_service_maker_droppper: unsafe fn(*const u8),
+            stop_caller: unsafe fn(*mut u8) -> VoidFuture,
+            dropper: unsafe fn(*mut u8),
+        ) -> VoidFuture;
+
+        fn stop(distr: &distributed) -> VoidFuture;
+    }
+}
+
+fn stop_caller<S: Service>(raw_service: *mut u8) -> VoidFuture {
+    VoidFuture::infallible_local(async move {
+        let stop_future = unsafe { (raw_service as *mut S).as_mut().unwrap().stop() };
+        Pin::from(stop_future).await;
+    })
+}
+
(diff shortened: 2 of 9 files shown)


kfernandez31:
There is currently an issue - running the tests with more than 2 logical cores results in the occasional segfault or an absurdly long wait time. I'm in the process of deducing what's the issue at hand.

seastar/src/distributed.cc
```diff
@@ -0,0 +1,58 @@
...
+{
+    _inner = raw_service_maker_caller(raw_service_maker);
+}
+
+seastar::future<> rust_service::stop() {
+    co_await _stop_caller(_inner);
+}
+
+rust_service::~rust_service()
+{
```
piodul: Nit: please put this curly brace on the previous line

seastar/src/distributed.rs
```diff
@@ -0,0 +1,537 @@
...
+                    distr.as_ref().unwrap(),
+                    boxed_raw_service_maker,
+                    raw_service_maker_caller,
+                    stop_caller,
+                    dropper,
+                )
+            }
+        };
+
+        let maker_dropper = Dropper::new(boxed_raw_service_maker, raw_service_maker_dropper);
```
piodul: Unfortunately, this doesn't solve the problem of "service_maker" being dropped too early. For example, the future returned by `start_inner` can be dropped immediately before "service_maker" has a chance to be called at all.

I think you should just transfer ownership of the service maker to the c++ side, and let `ffi::start`/`ffi::start_single` destroy it when it is appropriate.

seastar/src/distributed.rs
```diff
@@ -0,0 +1,537 @@
...
+    }
+}
+
+impl<S: Service> Deref for Distributed<S> {
+    type Target = S;
+    fn deref(&self) -> &S {
+        let local = ffi::local(self._inner.as_ref().unwrap());
+        unsafe { &*(local as *const S) }
+    }
+}
```
piodul: This should be an explicit `local()` method instead of a `Deref` impl. `Deref` might be confusing because it is usually implemented for "smart pointer-like" types. I don't think that you can think about `distributed` in such way, mainly because you will get references to different objects depending on the shard you are performing the dereference on.

seastar/src/distributed.rs
```diff
@@ -0,0 +1,537 @@
...
+        //         // You can't use `async` in a `Drop` implementation,
+        //         // so you need to block the current thread to wait for the `stop` method to complete.
+        //         futures::executor::block_on(async {
+        //             self.stop().await;
+        //         });
+        //     }
+        // }
+        //```
+        // If we decide to do so, it'd make sense to change the signature of `stop` to `async fn stop(&mut self)`.
+        // This however would produce an inconsistency, as the `start*` methods operate on interior mutability (take by `&self`).
```
piodul: No, we musn't block in Drop. Doing this would block the shard's reactor and would prevent other tasks from running - which would cause a deadlock if `stop()` spawns some tasks and then waits for them.

piodul (changes requested):

seastar/src/distributed.rs
kfernandez31: Sure

seastar/src/distributed.rs
kfernandez31: Can I send a clone of the maker to each shard and have each `rust_service` point to it and its dropper, and then free the maker's memory once it constructs itself?

piodul: Sounds like it would work, but now you have to also teach C++ how to clone the maker. I don't think it's necessary to clone the maker, just make sure to destroy it after `distributed::start` finishes. So:

```c++
VoidFuture start(
    const distributed& distr,
    uint8_t* raw_service_maker,
    rust::Fn<uint8_t*(uint8_t*)> raw_service_maker_caller,
    /* raw_service_maker_dropper */,
    rust::Fn<VoidFuture(uint8_t*)> stop_caller,
    rust::Fn<void(uint8_t*)> dropper
) {
    // seastar::defer just runs the specified lambda in its destructor, so that it is cleaned up regardless of whether distributed::start throws an exception or not
    auto cleanup_maker = seastar::defer([&] {
        raw_service_maker_dropper(raw_service_maker);
    });
    co_await const_cast<distributed&>(distr).start(raw_service_maker, raw_service_maker_caller, /*raw_service_maker_dropper,*/ stop_caller, dropper);
}
```

kfernandez31: Right. I haven't used `defer` before, but from your snippet, and Seastar's README I presume this works as expected.

kfernandez31:
v2: resolved @piodul 's comments

seastar/src/ffi_utils.rs
```diff
@@ -15,6 +15,28 @@ where
     fn_once_caller::<Func, Ret>
 }
 
+/// Convert the pointer to Box<Func> and then call it but don't consume.
+fn fn_caller<Func, Ret>(raw_func: *mut u8) -> Ret
+where
+    Func: Fn() -> Ret,
+{
+    unsafe {
+        let boxed_func = Box::from_raw(raw_func as *mut Func);
```
piodul: I'm not sure if it is legal to temporarily convert to a Box and then release. If a panic happens during `boxed_func()`, then the panic will cause the `boxed_func` to be dropped, which is not what we want.

I think you can just convert `raw_func` to `*mut Func`, then `&Func` and then just call it.

seastar/src/ffi_utils.rs
```diff
@@ -15,6 +15,28 @@ where
...
+        let ret = boxed_func();
+        std::ptr::swap(raw_func, Box::into_raw(boxed_func) as *mut u8);
```
piodul: What is this line supposed to do?

https://doc.rust-lang.org/std/ptr/fn.swap.html

This line just swaps two bytes.

kfernandez31: Well, putting the function in a `Box` meant that the `Box` now made sure to drop it, but we didn't want that so that `raw_func` would still point to valid data. So I extracted the data out of the `Box` and wanted to put it back, which I opted to do with `swap`

seastar/src/ffi_utils.rs
kfernandez31: Yeah this makes sense.

seastar/src/ffi_utils.rs
kfernandez31: This won't be needed as I followed your advice in the other thread, but tell me if my reasoning was flawed.

kfernandez31:
v3: fixed 2 more issues addressed by @piodul

seastar/src/ffi_utils.rs
piodul: `Box::from_raw` creates a Box which assumes ownership of the pointer you give to it. That Box will not move memory, just use the same pointer.

`Box::into_raw` just forgets about the `Box` wrapper and returns the pointer it contains. It does not "extract" any data behind the pointer.

`std::ptr::swap` takes two pointers and swaps the memory that they point to, based on their type. Given that they are of type `*mut u8` here, the function will just take the bytes pointed by them and swap them together. More specifically, `raw_func` and `Box::into_raw(boxed_func) as *mut u8` should be equal so the function will not do anything useful.

It would make more sense to call `std::ptr::swap` on two pointers of type `*mut Func`. Currently, you risk triggering undefined behavior if `Func` is a ZST (zero-sized type).

Besides, I don't think you should really touch the memory behind `raw_func`. `raw_func` should be a const pointer, as the function behind `raw_func` implements `Fn()` and possibly `Sync`, so multiple threads/shards might use it at the same time.

kfernandez31: I see, so I could have just done `let _ = Box::from_raw(boxed_func)`.
Anyways, I changed `raw_func` to be a pointer to `const`, but this required me to write a sort of doubtfully pretty `get_dropper_const`. Take a look and see if we're content with that.

piodul (approved):

PR #41, opened 2023-03-28
Expose basic file I/O

### LINKED ISSUE #39: Expose basic file I/O
We would like to allow for I/O operations on files.

Proposed interface:

```rust
use std::io;
use std::path::Path;

// A buffer that stores/receives data for I/O operations.
// Its contents are aligned in memory up to 512 bytes.
// read_dma and write_dma require memory to be aligned.
pub struct DmaBuffer {
    chunks: Vec<DmaChunk>,
}

#[repr(align(512))]
struct DmaChunk([u8; 4096]);

// Interface can be modelled after std::fs::OpenOptions
// (a builder)
// https://doc.rust-lang.org/std/fs/struct.OpenOptions.html
pub struct OpenOptions {
    // ...
}

pub struct File {
    // ...
}

// Ref: seastar/core/file.hh
impl File {
    // Read some bytes at given position.
    //
    // The function does not overwrite existing data in the DmaBuffer,
    // but rather reads the data into the free capacity of the internal Vec.
    //
    // Returns the number of bytes read and the original buffer.
    pub async fn read_dma(
        &self,
        buffer: DmaBuffer,
        pos: u64,
    ) -> Result<(usize, DmaBuffer), io::Error>;

    // Writes some bytes at given position.
    //
    // Returns (bytes_written, original_buffer) on success.
    pub async fn write_dma(
        &self,
        buffer: DmaBuffer,
        pos: u64,
    ) -> Result<(usize, DmaBuffer)>;

    pub async fn flush(&self) -> Result<(), io::Error>;
    pub async fn close(&self) -> Result<(), io::Error>;

    pub async fn size(&self) -> Result<u64, io::Error>;
}
```

Caveats:

- It should be safe for all methods of `File` to take `self` by immutable reference.
- Be careful about the buffer lifetime. In Rust, a future can be dropped before the operation it waits on completes. More specifically, it is legal to drop a future which waits until a read or a write finishes - it's important to make sure that the buffer is not dropped until the operation completes. Because of that, the buffers are passed by value (similarly to the design of tokio_uring).
- `io_priority_class` can be skipped for now.
- Nice to have, but not necessary: we could use the `io_intent` to cancel an outstanding read/write operation in case the Rust future is dropped. This does not invalidate the first point, we still must make sure that the buffer lives until the futures returned by `dma_write`/`dma_read` resolve.

grzenow4:
Fixes: #39
Depends on: #24 #27

This PR introduces structs `OpenOptions` and `File`
```rust
#[derive(Clone)]
pub struct OpenOptions {
    read: bool,
    write: bool,
    create: bool,
}

pub struct File {
    inner: UniquePtr<file_t>,
}
```
which allow for basic I/O operations such as read and write to a file. All the introduced methods are:
- `read_dma`
- `write_dma`
- `flush`
- `close`
- `size`

Reading and writing to a file is done via `DmaBuffer` struct:
```rust
const ALIGN: usize = 512;
const CHUNK_SIZE: usize = 4096;

pub struct DmaBuffer {
    buffer: *mut u8,
    size: usize,
}
```

### DIFF
--- a/seastar/src/file.rs
+++ b/seastar/src/file.rs
@@ -1,0 +1,427 @@
+use crate::assert_runtime_is_running;
+use cxx::UniquePtr;
+use ffi::*;
+use std::alloc::{self, Layout};
+use std::io;
+use std::ops::{Deref, Index, IndexMut};
+use std::path::Path;
+
+#[cxx::bridge]
+mod ffi {
+    extern "Rust" {
+        type OpenOptions;
+
+        pub fn get_read(&self) -> bool;
+        pub fn get_write(&self) -> bool;
+        pub fn get_create(&self) -> bool;
+    }
+
+    #[namespace = "seastar_ffi"]
+    unsafe extern "C++" {
+        type VoidFuture = crate::cxx_async_futures::VoidFuture;
+        type IntFuture = crate::cxx_async_futures::IntFuture;
+    }
+
+    #[namespace = "seastar_ffi::file"]
+    unsafe extern "C++" {
+        include!("seastar/src/file.hh");
+
+        type file_t;
+
+        fn open_dma(file: &mut UniquePtr<file_t>, name: &str, opts: &OpenOptions) -> VoidFuture;
+
+        unsafe fn read_dma(
+            file: &UniquePtr<file_t>,
+            buffer: *mut u8,
+            size: u64,
+            pos: u64,
+        ) -> IntFuture;
+
+        unsafe fn write_dma(
+            file: &UniquePtr<file_t>,
+            buffer: *mut u8,
+            size: u64,
+            pos: u64,
+        ) -> IntFuture;
+
+        fn flush(file: &UniquePtr<file_t>) -> VoidFuture;
+
+        fn close(file: &UniquePtr<file_t>) -> VoidFuture;
+
+        fn size(file: &UniquePtr<file_t>) -> IntFuture;
+    }
+}
+
+const ALIGN: usize = 512;
+const CHUNK_SIZE: usize = 4096;
+
+/// A buffer that stores/receives data for I/O operations.
+/// Its contents are aligned in memory up to 512 bytes.
+/// read_dma and write_dma require memory to be aligned.
+pub struct DmaBuffer {
+    buffer: *mut u8,
+    size: usize,
+}
+
+impl Deref for DmaBuffer {
+    type Target = [u8];
+
+    fn deref(&self) -> &Self::Target {
+        self.as_slice()
+    }
+}
+
+impl Drop for DmaBuffer {
+    fn drop(&mut self) {
+        let layout = Layout::from_size_align(self.size, ALIGN).unwrap();
+        unsafe {
+            alloc::dealloc(self.buffer, layout);
+        }
+    }
+}
+
+impl Index<usize> for DmaBuffer {
+    type Output = u8;
+
+    fn index(&self, idx: usize) -> &Self::Output {
+        self.as_slice().index(idx)
+    }
+}
+
+impl IndexMut<usize> for DmaBuffer {
+    fn index_mut(&mut self, idx: usize) -> &mut Self::Output {
+        self.as_mut_slice().index_mut(idx)
+    }
+}
+
+impl DmaBuffer {
+    pub fn from_slice(bytes: &[u8]) -> Self {
+        let size = bytes.len();
+        assert!(size % CHUNK_SIZE == 0);
+        let layout = Layout::from_size_align(size, ALIGN).unwrap();
+        unsafe {
+            let buffer = alloc::alloc_zeroed(layout);
+            let slice = std::slice::from_raw_parts_mut(buffer, size);
+            slice.copy_from_slice(bytes);
+            Self { buffer, size }
+        }
+    }
+
+    pub fn copy_from_slice(&mut self, bytes: &[u8]) -> &mut Self {
+        self.as_mut_slice().copy_from_slice(bytes);
+        self
+    }
+
+    pub fn as_mut_slice(&mut self) -> &mut [u8] {
+        unsafe { std::slice::from_raw_parts_mut(self.buffer, self.size) }
+    }
+
+    pub fn as_slice(&self) -> &[u8] {
+        unsafe { std::slice::from_raw_parts(self.buffer, self.size) }
+    }
+}
+
+/// Interface modelled after std::fs::OpenOptions.
+/// It is a builder for the File.
+#[derive(Clone)]
+pub struct OpenOptions {
+    read: bool,
+    write: bool,
+    create: bool,
+}
+
+impl Default for OpenOptions {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+impl OpenOptions {
+    /// Creates a new `OpenOptions` struct with no flag set.
+    pub fn new() -> Self {
+        Self {
+            read: false,
+            write: false,
+            create: false,
+        }
+    }
+
+    /// Sets a flag `read` which allows to read from a file.
+    pub fn read(&mut self, flag: bool) -> &mut Self {
+        self.read = flag;
+        self
+    }
+
+    /// Sets a flag `write` which allows to write to a file.
+    pub fn write(&mut self, flag: bool) -> &mut Self {
+        self.write = flag;
+        self
+    }
+
+    /// Sets a flag `create` which allows to create a new file.
+    pub fn create(&mut self, flag: bool) -> &mut Self {
+        self.create = flag;
+        self
+    }
+
+    /// Getter for a `read` flag.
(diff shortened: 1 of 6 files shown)


seastar/src/file.rs
```diff
@@ -0,0 +1,298 @@
...
+        fn flush(file: &UniquePtr<file_t>) -> VoidFuture;
+
+        fn close(file: &UniquePtr<file_t>) -> VoidFuture;
+
+        fn size(file: &UniquePtr<file_t>) -> IntFuture;
+    }
+}
+
+#[repr(align(512))]
+struct DmaChunk([u8; 4096]);
```
Lorak-mmk: This is a constant appearing in more than one place - please extract it to global const.

seastar/src/file.rs
```diff
@@ -0,0 +1,298 @@
...
+    pub async fn write_dma(
+        &self,
+        buffer: DmaBuffer,
+        pos: u64,
+    ) -> Result<(usize, DmaBuffer), io::Error> {
+        assert_runtime_is_running();
+        let buf_ptr = buffer.chunks.as_ptr() as *mut u8;
+        let size = buffer.chunks.len() as u64;
+        unsafe {
+            let fut = write_dma(&self.inner, buf_ptr, size * 4096, pos);
```
Lorak-mmk: This is may be just a subjective opinion, but I feel this multiplication should me move to `size` declaration.

seastar/src/file.rs
```diff
@@ -0,0 +1,298 @@
...
+            .unwrap();
+        let file = OpenOptions::new()
+            .read(true)
+            .open("example-read")
+            .await
+            .unwrap();
+        let chunks = vec![DmaChunk([0u8; 4096])];
+        let buffer = DmaBuffer { chunks };
+        let res = file.read_dma(buffer, 0).await.unwrap();
+        file.flush().await.unwrap();
```
Lorak-mmk: Do we need flush if we didn't write anything to the file?

Lorak-mmk (changes requested):
- There should be some tests that read / write more than 1 chunk
- Currently, the interface seems quite difficult to use, due to need to use `DmaBuffer`. I think some helper methods, that allow user to:
    - create new `DmaBuffer` from vector of bytes / string
    - copy content of vector of bytes / string to existing `DmaBuffer` 
    - cast `DmaBuffer` to a slice of it's contents `as_mut_slice(&mut self) -> &mut [u8]` and `as_slice(&self) -> &[u8]`.
    - It could also implement `Deref<Target=[u8]>`, `Index`, `IndexMut`

seastar/src/file.rs
```diff
@@ -0,0 +1,407 @@
...
+        self.buffer.as_slice()
+    }
+
+    // We need this function to also update `buffer` field after the call `read_dma()`.
+    fn update(mut self) -> Self {
+        for (i, chunk) in self.chunks.iter().enumerate() {
+            self.buffer[i * CHUNK_SIZE..i * CHUNK_SIZE + CHUNK_SIZE].copy_from_slice(&chunk.0);
+        }
+        self
+    }
```
piodul: It looks like you are trying to solve the problem of casting the Vec of DmaChunks to a byte slice by keeping a separate Vec<u8> and copying data there as needed. That's really not great - we would like to reduce copying to minimum.

If it's hard to make the `Vec<DmaChunk>` to work then perhaps we could try something else. I found that Rust exposes functions for allocating data directly.

- Use `std::alloc::alloc_zeroed` to create a buffer with correct size and alignment (see the `Layout` struct): https://doc.rust-lang.org/nightly/std/alloc/fn.alloc_zeroed.html
- Use `std::alloc::dealloc` in `impl Drop` to deallocate the buffer: https://doc.rust-lang.org/nightly/std/alloc/fn.dealloc.html
- Use `std::slice::from_raw_parts` to turn the pointer into a proper slice: https://doc.rust-lang.org/std/slice/fn.from_raw_parts.html

seastar/src/file.rs
```diff
@@ -0,0 +1,407 @@
...
+            .unwrap()
+            .read_to_end(&mut line)
+            .unwrap();
+        let bytes = res.1.as_slice();
+        assert_eq!(bytes, line.as_slice());
+    }
+
+    #[seastar::test]
+    async fn test_file_write_dma_big() {
+        let v = (0..10000).map(|_| rand::random::<u8>()).collect();
```
piodul: The size of `v` is not aligned to 4 kilobytes. `DmaBuffer::new()` will truncate some of the bytes at the end and will only write 8 kilobytes to disk.

seastar/src/file.rs
```diff
@@ -0,0 +1,407 @@
...
+    use crate as seastar;
+    use std::io::{Read, Write};
+
+    #[seastar::test]
+    async fn test_file_read_dma() {
+        let msg = b"I <3 seastar!";
+        std::fs::OpenOptions::new()
+            .create(true)
+            .write(true)
+            .open("example-read")
```
piodul: It would be really nice if the tests generated random file names and put everything to the temporary directory. Otherwise the tests might behave differently if you run them for the second time.

See here: https://doc.rust-lang.org/std/env/fn.temp_dir.html

seastar/src/file.rs
```diff
@@ -0,0 +1,407 @@
...
+        unsafe {
+            let fut = write_dma(&self.inner, buf_ptr, size, pos);
+            match fut.await {
+                Ok(res) => Ok((res as usize, buffer)),
+                Err(e) => Err(io::Error::new(io::ErrorKind::Other, e)),
+            }
+        }
+    }
+
+    pub async fn flush(&self) -> Result<(), io::Error> {
```
piodul, line 247: Please add docstrings to the remaining methods.

seastar/src/file.rs
```diff
@@ -0,0 +1,407 @@
...
+    create: bool,
+}
+
+impl Default for OpenOptions {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+impl OpenOptions {
```
piodul, line 139: Please add docstrings for the functions/methods.

seastar/src/file.rs
piodul: Besides, I don't think `from_vec` should truncate. I would prefer if it panicked or returned an error if the vec's length is not a multiple of 4kb. Silently dropping data like this really not cool.

piodul (changes requested):

seastar/src/file.rs
grzenow4: It was intentional to make a such a test. I thought that `DmaBuffer` should truncate excessive bytes or fill a chunk with `0u8` if vec is too small. But having 4kB alignment assert sounds better.

seastar/src/file.rs
```diff
@@ -0,0 +1,433 @@
...
+    }
+
+    pub fn copy_vec(&mut self, bytes: Vec<u8>) -> &mut Self {
+        self.as_mut_slice().copy_from_slice(bytes.as_slice());
+        self
+    }
+
+    pub fn copy_string(&mut self, bytes: String) -> &mut Self {
+        self.copy_vec(bytes.into_bytes())
+    }
```
piodul: Three issues with those functions:

- They take the arguments by value, which means that they are dropped at the end of those methods. This is absolutely unnecessary and those should be taken by reference - even better, they should take a bytes slice (`&'a [u8]`).
- I don't think we need to expose a dedicated function for copying from a string. It's easy to convert a str or String to a slice of bytes (via `as_bytes()`).
- The name is not clear. "copy_vec", but do we copy to the vec, from the vec? The function should ultimately be converted to take a bytes slice, so `copy_from_slice` would be the best option as it uses a convention established in the standard library.

seastar/src/file.rs
```diff
@@ -0,0 +1,433 @@
...
+}
+
+impl IndexMut<usize> for DmaBuffer {
+    fn index_mut(&mut self, idx: usize) -> &mut Self::Output {
+        self.as_mut_slice().index_mut(idx)
+    }
+}
+
+impl DmaBuffer {
+    pub fn from_vec(bytes: Vec<u8>) -> Self {
```
piodul: This function should take a slice of bytes.

seastar/src/file.rs
```diff
@@ -0,0 +1,433 @@
...
+}
+
+impl Default for OpenOptions {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+impl OpenOptions {
+    // Creates a new `OpenOptions` struct with no flag set.
```
piodul: The comments that you added won't show up as docstrings. Docstrings use three slashes instead of two. If you want to check that your docstrings are rendered properly, you can use `cargo doc` to generate the documentation.

piodul (changes requested):
Needs rebase.

Lorak-mmk (approved):

piodul (approved):

PR #42, opened 2023-03-29
logger: add Rust interface to seastar loggers

piodul:
Adds an API that wraps seastar loggers.

A logger can be instantiated on stack, but it can also be defined globally, with the help of the `ctor` crate:

    #[ctor::ctor]
    static MLOGGER: Logger = Logger::new("main");

Log messages can be printed with the help of macros that follow a syntax similar to `println!` and its friends:

    seastar::info!(MLOGGER, "Going to frobnicate {} blubs", num_blubs);

### DIFF
--- a/seastar/src/logger.rs
+++ b/seastar/src/logger.rs
@@ -1,0 +1,248 @@
+use std::fmt::Arguments;
+use std::pin::Pin;
+
+use cxx::UniquePtr;
+
+#[cxx::bridge(namespace = "seastar_ffi::logger")]
+mod ffi {
+    extern "Rust" {
+        type FormatCtx<'a>;
+        fn write_log_line(writer: Pin<&mut log_writer>, ctx: &FormatCtx<'_>);
+    }
+
+    unsafe extern "C++" {
+        include!("seastar/src/logger.hh");
+
+        type log_writer;
+        fn write(self: Pin<&mut log_writer>, data: &[u8]);
+
+        type logger;
+        fn new_logger(name: &str) -> UniquePtr<logger>;
+        fn log(l: &logger, level: u32, ctx: &FormatCtx<'_>);
+    }
+}
+
+/// Internal, do not use.
+// For some reason, cxx requires this to be public.
+#[doc(hidden)]
+pub struct FormatCtx<'a> {
+    args: Arguments<'a>,
+}
+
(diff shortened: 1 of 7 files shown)


seastar/src/logger.rs
```diff
@@ -0,0 +1,248 @@
...
+
+    unsafe extern "C++" {
+        include!("seastar/src/logger.hh");
+
+        type log_writer;
+        fn write(self: Pin<&mut log_writer>, data: &[u8]);
+
+        type logger;
+        fn new_logger(name: &str) -> UniquePtr<logger>;
+        fn log<'a>(l: &logger, level: u32, ctx: &FormatCtx<'_>);
```
kfernandez31, line 21: The lifetime `'a` isn't used in the function definition. `'_` should suffice.

kfernandez31: The lifetime `'a` isn't used in the function definition. `'_` should suffice.

seastar/src/logger.rs
```diff
@@ -0,0 +1,248 @@
...
+
+/// Emits an `error` level message with given logger.
+///
+/// Equivalent to calling [`log!`](crate::log!) with `error` level.
+#[macro_export]
+macro_rules! error {
+    ($logger:expr, $($arg:tt),*) => {{
+        $logger.error(std::format_args!($($arg),*))
+    }};
+}
```
kfernandez31, line 248: Maybe we could have some tests that capture `stderr` and assert whether it contains the expected logs?

piodul: I'd very much like to, but I'm not sure how to do it. It appears that there is a function in seastar that allows substituting a custom ostream in place of the default std::cerr, but the substitution is global, which won't interact well with multithreaded tests.

Issues in log output is something that is very visible and detected very quickly by developers, so I think I'd rather skip adding such tests for now (unless you have a good idea on how to solve the issue of the global).

seastar/src/logger.rs
piodul: Actually, the lifetime parameter could be completely removed, but thanks for pointing the issue out.

piodul:
@kfernandez31 Sorry for taking such a long time. Could you take one more look?

seastar/src/logger.rs
kfernandez31: Yeah, understandable.

kfernandez31:
LGTM

Lorak-mmk (approved):

PR #45, opened 2023-04-18
key-value store example

patjed41:
Depends on #46

This PR introduces a key-value store demo of `seaster-rs`.

The description is in `seastar-rs/examples/key-value-store/README.md`. I think describing this demo is a good idea because reading its code is one of the primary ways for users to learn how to use our crate. That's why I have written a quite precise README file.

I would like to point out some technical issues regarding the demo and its development:
- When I was trying to use `seastar::Distributed<S>`, it turned out that some things must be fixed or improved there. @kfernandez31 made all improvements necessary to make the example work, but he has not finished yet. He will make a new PR when he finishes. So the first three commits of this PR containing his changes should not be discussed here.
- For now the field `db` of the struct `Server` has type`RefCell<HashMap<String, String>>`. `RefCell` is necessary because `seastar::Distributed<S>::map_single` requires a const reference of the service. If this changes, I will remove `RefCell`.
- The two issues described above are the reason I marked this PR as a draft. Some parts of `seastar::Distributed<S>` can change and it will affect this PR.
- The example is a completely separate crate because `build.rs` is necessary. If I'm not wrong, it's impossible to create such examples in the `seastar/examples` directory.
- Currently, `listen` calls ending with an error are not handled in Rust. Underlying `seastar::listen` ends the program with a stack trace. It happens, for example, if port 5555 is already in use. I'm not sure if we should handle it in a different way.
- The whole `requests` module is almost an exact copy-paste from the [Tokio server](https://github.com/patjed41/key-value-store/blob/master/src/request_parsing.rs) done by me at the beginning of the ZPP. It has already passed a review and unit tests so I think it is safe to assume that it just works.

### DIFF
--- a/examples/key-value-store/README.md
+++ b/examples/key-value-store/README.md
@@ -1,0 +1,57 @@
+# key-value store
+
+This crate is an example of using the crate `seastar-rs`. Its two main objectives are:
+1. To demonstrate how to use some of Seastar's functionalities exposed by `seastar-rs` from C++ to Rust. This example covers [`seastar::main`](https://github.com/zpp-2022-rust-seastar/seastar-rs/blob/main/seastar-macros/src/lib.rs) macro and [`seastar::Distributed<S>`](https://github.com/zpp-2022-rust-seastar/seastar-rs/blob/main/seastar/src/distributed.rs).
+2. To show how to integrate the code written using `seastar-rs `with the already existing C++ code. In this example, a part of Seastar's network API is exposed to Rust using the `cxx` crate and integrated with the rest of the Rust code.
+
+## Use
+
+To run this example, execute the following command in the `seastar-rs` directory:
+```
+RUSTFLAGS="-C link-arg=-fuse-ld=lld" RUSTDOCFLAGS="-C link-arg=-fuse-ld=lld" cargo run --bin key-value-store
+```
+
+If you work on a machine with more than 30 cores, running the example can end up in a deadlock due to a bug in the `cxx-async` crate. In such a case, the `taskset` command might help. For instance, you could run the example using 16 cores like this:
+
+```
+RUSTFLAGS="-C link-arg=-fuse-ld=lld" RUSTDOCFLAGS="-C link-arg=-fuse-ld=lld" taskset -c 0-15 cargo run --bin key-value-store
+```
+
+If you want to connect to the server, you could use `netcat`:
+```
+nc <address> 5555
+```
+
+## Server
+
+This example is an implementation of a simple key-value store server. The server accepts TCP connections on port 5555 and serves client requests. Each request is an ASCII string.
+
+### Requests
+
+The server accepts two kinds of requests - STORE and LOAD.
+
+STORE requests:
+  - have a form `STORE$key$value\n` where `key` and `value` can be a string of any length (even empty) containing only lowercase letters of the English alphabet,
+  - after receiving such a request, the server puts `value` under `key` in its memoty,
+  - if `key` is already present in the server's memory, `value` overwrites the old value,
+  - the server responds with `DONE\n`.
+
+LOAD requests:
+  - have a form `LOAD$key\n` where `key` can be a string of any length containing only lowercase letters of the English alphabet,
+  - after receiving such a request, if `key` is present in the server's memory, the server sends a reposone of a form `FOUND$value\n` where `value` is the value under `key`,
+  - if `key` is not present in the server's memory, the server sends a response `NOTFOUND\n`.
+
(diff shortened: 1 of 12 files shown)


examples/key-value-store/README.md
```diff
@@ -0,0 +1,57 @@
+# key-value store
+
+This crate is an example of using the crate `seastar-rs`. Its two main objectives are:
+1. To demonstrate how to use some of Seastar's functionalities exposed by `seastar-rs from` C++ to Rust. This example covers [`seastar::main`](https://github.com/zpp-2022-rust-seastar/seastar-rs/blob/main/seastar-macros/src/lib.rs) macro and [`seastar::Distributed<S>`](https://github.com/zpp-2022-rust-seastar/seastar-rs/blob/main/seastar/src/distributed.rs).
```
Lorak-mmk:
```suggestion
1. To demonstrate how to use some of Seastar's functionalities exposed by `seastar-rs` from C++ to Rust. This example covers [`seastar::main`](https://github.com/zpp-2022-rust-seastar/seastar-rs/blob/main/seastar-macros/src/lib.rs) macro and [`seastar::Distributed<S>`](https://github.com/zpp-2022-rust-seastar/seastar-rs/blob/main/seastar/src/distributed.rs).
```

examples/key-value-store/src/net_ffi.cc
```diff
@@ -0,0 +1,55 @@
...
+    co_return rust::String(buffer.begin(), buffer.size());
+}
+
+VoidFuture write(
+    const std::unique_ptr<output_stream>& output,
+    const rust::str msg
+) {
+    co_await output->write(seastar::sstring(msg.begin(), msg.size()));
+    co_await output->flush();
+}
```
Lorak-mmk: Flush should be a separate method.

examples/key-value-store/src/net_ffi.rs
```diff
@@ -0,0 +1,64 @@
...
+unsafe impl Future for VoidFuture {
+    type Output = ();
+}
+
+#[cxx_async::bridge(namespace = net_ffi)]
+unsafe impl Future for StringFuture {
+    type Output = String;
+}
+
+unsafe impl Send for ConnectedSocket {}
```
Lorak-mmk: This requires a comment explaining why it's sound to implement Send here.

Lorak-mmk (changes requested):

examples/key-value-store/src/net_ffi.cc
patjed41: I made a new function `net_ffi:flush_output`. This change not only makes sense stylistically but also enables a tiny improvement. There is no reason to flush the output stream after every write. One message from a client can contain many requests. Flushing after responding to all of them is sufficient. That is why the new `net_ffi:flush_output` is called after responding to all requests in the `ShardedServer::handle_connection` method.

examples/key-value-store/src/net_ffi.rs
patjed41: I got rid of `unsafe impl Send for ConnectedSocket {}`. `Disributed::map_single` required this. However, there is no reason to require the `Send` trait when `map_single` is called on the current shard. That is why the function `Disributed::map_current` has been created in #46, and I used it here.

examples/key-value-store/src/net_ffi.cc
Lorak-mmk: What if a client send a request, waits for response and only then sends another one? Isn't flush after each request helpful in this scenario?

Lorak-mmk:
One more thing, can you change the protocol slightly, so that each request ends with dollar and newline instead of dollar? It will make it easier to send requests manually, should anyone try to use this example.

examples/key-value-store/src/net_ffi.cc
patjed41: In such a case, everything would work fine. The server would instantly flush the output stream after each request. The difference after the change is that the server flushes after processing a single message (a single `read`) and not after processing a single request. Usually, there is one request per message, but the client can send multiple requests in a single message. Then, this change matters.

I still think it always works fine, but maybe I am wrong, and there is some much weirder scenario when this change is detrimental. It also makes the code less understandable. I think code that for sure works and is easier to understand is much better in a demo, so I decided to undo this change.

patjed41:
> One more thing, can you change the protocol slightly, so that each request ends with dollar and newline instead of dollar? It will make it easier to send requests manually, should anyone try to use this example.

Sure, I changed it. Should I also change responses (`FOUND$` etc.) in the same way? It would make using the example with programs like netcat easier. Every request and response would be in a separate line. However, I think it would be not so good for non-manual use.

Lorak-mmk:
Yeah, you can change responses too

Lorak-mmk (approved):

PR #46, opened 2023-04-24
Improvements for Distributed

kfernandez31:
This branch introduces:
- missing safety checks (kind of a big deal),
- more types of maps for a more ergonomic API. The ones that existed before now have their `_mut` counterparts, which allow for regular mutability, instead of interior mutability. Note, however, that `stop` still requires interior mutability, should you want to alter the service's state.
- the much-awaited functionality of `peering_sharded_service`, through which instances can communicate and delegate work.

### DIFF
--- a/seastar/src/distributed.rs
+++ b/seastar/src/distributed.rs
@@ -1,6 +1,6 @@
 use crate::{
     cxx_async_local_future::IntoCxxAsyncLocalFuture,
-    ffi_utils::{get_dropper_const, get_dropper_noarg, get_fn_caller},
+    ffi_utils::{get_dropper_const, get_dropper_noarg, get_fn_caller, PtrWrapper},
     get_count,
     submit_to::submit_to,
     this_shard_id,
@@ -89,6 +89,18 @@ pub trait Service {
     fn stop(&self) -> Box<dyn Future<Output = ()>> {
         Box::new(async {})
     }
+}
+
+/// An object on which `Distributed`'s mapping functions operate.
+///
+/// It provides access to the local instance of a service,
+/// as well as its container, which can be used to communicate with other shards.
+pub struct PeeringShardedService<'a, S>
+where
+    S: Service,
+{
+    pub instance: &'a S,
+    pub container: &'a Distributed<S>,
 }
 
 /// A service distributed amongst all shards of a Seastar app.
@@ -265,13 +277,40 @@ impl<S: Service> Distributed<S> {
         ffi::stop(self._inner.as_ref().unwrap()).await.unwrap();
     }
 
+    fn submit_to<'a, Func, Fut, Ret>(
+        &'a self,
+        shard_id: u32,
+        func: Func,
+        container: PtrWrapper,
+    ) -> impl Future<Output = Ret>
+    where
+        Func: FnOnce(PeeringShardedService<'a, S>) -> Fut + Send + 'static,
+        Fut: Future<Output = Ret>,
+        Ret: Send + 'static,
+    {
+        crate::assert_runtime_is_running();
+
+        let distr = self._inner.clone();
+        submit_to(shard_id, || async move {
+            let instance = ffi::local(distr.as_ref().unwrap());
+            let instance: &S = unsafe { &*(instance as *const S) };
+            let _ = &container; // this is to avoid a partial move of the pointer
+            let container = unsafe { &*(container.as_ptr_mut() as *const Distributed<S>) };
+            let pss = PeeringShardedService {
+                instance,
+                container,
+            };
+            func(pss).await
+        })
+    }
+
     fn map_selected<'a, Func, Fut, Ret, I>(
         &'a self,
         func: Func,
         shards: I,
     ) -> Vec<impl Future<Output = Ret>>
     where
-        Func: Fn(&'a S) -> Fut + Send + Clone + 'static,
+        Func: FnOnce(PeeringShardedService<'a, S>) -> Fut + Send + Clone + 'static,
         Fut: Future<Output = Ret>,
         Ret: Send + 'static,
         I: IntoIterator<Item = u32>,
@@ -280,12 +319,21 @@ impl<S: Service> Distributed<S> {
 
         shards
             .into_iter()
-            .map(|shard| (shard, func.clone(), self._inner.clone()))
-            .map(|(shard, func, inner)| async move {
+            .map(|shard| {
+                let distr = unsafe { PtrWrapper::new(self as *const Distributed<S> as _) };
+                (shard, func.clone(), self._inner.clone(), distr)
+            })
+            .map(|(shard, func, inner, distr)| async move {
                 submit_to(shard, || async move {
                     let local = ffi::local(inner.as_ref().unwrap());
                     let local = unsafe { &*(local as *const S) };
-                    func(local).await
+                    let _ = &distr; // this is to avoid a partial move of the pointer
+                    let distr = unsafe { &*(distr.as_ptr_mut() as *const Distributed<S>) };
+                    let pss = PeeringShardedService {
+                        instance: local,
+                        container: distr,
+                    };
+                    func(pss).await
                 })
                 .await
             })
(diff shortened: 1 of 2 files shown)


seastar/src/distributed.rs
```diff
@@ -303,6 +311,33 @@ impl<S: Service> Distributed<S> {
         })
     }
 
+    fn submit_to_mut<'a, Func, Fut, Ret>(
+        &'a self,
```
piodul, line 340: This is very unsafe. This allows creating mutable reference from an immutable reference without any checks. If `submit_to_mut` is called twice in a row then it can result in two tasks being created with a `&mut` reference to the same `S`, and this is undefined behavior.

This can only work if this function takes self as `&'a mut self`.

seastar/src/distributed.rs
```diff
@@ -273,6 +275,8 @@ impl<S: Service> Distributed<S> {
         Ret: Send + 'static,
         I: IntoIterator<Item = u32>,
     {
+        crate::assert_runtime_is_running();
```
piodul, line 389: The `stop` method is also missing such a check, no?

seastar/src/ffi_utils.rs
```diff
@@ -66,3 +66,10 @@ pub const fn get_dropper_const<T>(_: &T) -> fn(*const u8) {
 pub const fn get_dropper_noarg<T>() -> fn(*mut u8) {
     dropper::<T>
 }
+
+/// A wrapper over a raw pointer that makes it passable between threads.
+/// Use this only if you know what you're doing.
+pub struct PtrWrapper(pub *mut u8);
+
+unsafe impl Send for PtrWrapper {}
+unsafe impl Sync for PtrWrapper {}
```
piodul: This API is unsafe but allows itself to be used from safe code. That shouldn't be possible - you should add appropriate protections.

For example:

- Only allow creating a `PtrWrapper` via `unsafe fn new(ptr: *mut u8)`
- Only accessing the inner pointer via `unsafe fn as_ptr_mut(&self) -> *mut u8`.

seastar/src/distributed.rs
```diff
@@ -276,6 +276,33 @@ impl<S: Service> Distributed<S> {
         ffi::stop(self._inner.as_ref().unwrap()).await.unwrap();
     }
 
+    fn submit_to<'a, Func, Fut, Ret>(
```
piodul, line 306: `make maps create instantly running futures`

How about a test?

seastar/src/distributed.rs
```diff
@@ -279,12 +291,21 @@ impl<S: Service> Distributed<S> {
...
+            .map(|shard| {
+                let distr = PtrWrapper(self as *const Distributed<S> as _);
+                (shard, func.clone(), self._inner.clone(), distr)
+            })
+            .map(|(shard, func, inner, distr)| async move {
                 submit_to(shard, || async move {
                     let local = ffi::local(inner.as_ref().unwrap());
                     let local = unsafe { &*(local as *const S) };
-                    func(local).await
+                    let _ = &distr; // this is to avoid a partial move of the pointer
```
piodul: Could you elaborate? I don't understand what a "partial move" means here.

piodul (changes requested):
The first three commits look OK, but I don't think `map_mut` is going to work like this.

First and foremost, you must ensure exclusive access to the shard-local instances of the service. Unless you ensure this dynamically (e.g. via synchronization, which I'm not sure is the responsibility of `Distributed`), if you take `Distributed` by immutable reference then you risk creating two mutable references to the same shard-local instances, which is undefined behavior. The `_mut` methods need to take the whole `Distributed` instance via mutable reference.

Another thing is the "Peering" part. Due to the problems highlighted above, I don't think `map_mut` and friends should provide access to the container because it would allow creating more references to the same shard-local instances.

seastar/src/distributed.rs
kfernandez31: Without that line, `rustc` performed a partial move on the tuple-struct's pointer field.

kfernandez31:
v2 - resolved @piodul 's comments apart from this one:
> Another thing is the "Peering" part. Due to the problems highlighted above, I don't think map_mut and friends should provide access to the container because it would allow creating more references to the same shard-local instances.

I'd like to know whether the security checks I added have addressed your concerns or should I go ahead and rescind the access to the container of mut-maps.

seastar/src/distributed.rs
```diff
@@ -291,11 +312,20 @@ impl<S: Service> Distributed<S> {
         crate::assert_runtime_is_running();
 
         let distr = self._inner.clone();
-        submit_to(shard_id, || async move {
+        let lock = self._locks[shard_id as usize].clone();
+        submit_to(shard_id, move || async move {
+            let lock = lock.try_read();
+            if lock.is_err() {
+                panic!(
+                    "Data race detected: can't lock instance {} for read-access",
```
piodul: The message is misleading - it's not a data race. Data race happens when two or more threads try to access the same data in an unsynchronized way. Here, the lock will always be taken by only one thread - so it's more like you prevent a violation of Rust's aliasing rules. Perhaps you could make the message more similar to what a `RefCell` generates when it detects a rules violation.

Anyway, a more useful error message would also mention possible reasons for why the panic happened so that the developer quickly understands what they did wrong.

seastar/src/distributed.rs
```diff
@@ -513,6 +733,7 @@ mod tests {
 
     #[seastar::test]
     async fn test_map_all() {
+        //TODO
```
piodul: Leftover TODO?

seastar/src/distributed.rs
```diff
@@ -525,8 +746,29 @@ mod tests {
         assert_eq!(2 * get_count(), counter.load(Ordering::SeqCst));
     }
 
+    #[seastar::test]
+    async fn test_map_all_mut() {
+        //TODO
```
piodul: Leftover TODO?

seastar/src/distributed.rs
```diff
@@ -525,8 +746,29 @@ mod tests {
...
+            .filter(|x| x.eq(&false))
+            .count();
+
+        assert_eq!(count, 0);
+        distr.stop().await;
+    }
+
     #[seastar::test]
     async fn test_map_others() {
+        //TODO
```
piodul: More unexplained todos...

seastar/src/distributed.rs
```diff
@@ -304,6 +334,42 @@ impl<S: Service> Distributed<S> {
         })
     }
 
+    fn submit_to_mut<'a, Func, Fut, Ret>(
+        &'a self, // this immutable reference is intentional
```
piodul: Why is it intentional? This comment piques reader's curiosity but doesn't satisfy it.

seastar/src/distributed.rs
```diff
@@ -100,15 +103,32 @@ where
...
+///
+/// # Panics
+/// One can only delegeate a "mutating map" on an instance if all other "mutating maps" on it have ceased.
+/// In other words, you must ensure sequential acquisition of mutable borrows to an instance.
+/// Failing to comply with this rule will lead to a `panic!`, much in the same way that breaking the rule for
+/// a `RefCell` would do.
 pub struct Distributed<S: Service> {
     _inner: SharedPtr<distributed>,
     _ty: PhantomData<S>,
+    _locks: Vec<Arc<RwLock<()>>>,
```
piodul, line 136: Seeing a synchronous `RwLock` here looks very scary because using synchronous locks in asynchronous code can easily lead to a deadlock. However, I understand that it was easier to use compared to `RefCell` because `RefCell` is not Sync/Send and it would cause lots of problems and the need to use unsafe. Please add a comment which explains what is the purpose of those locks.

seastar/src/distributed.rs
kfernandez31: yup, I definetely missed them :p

seastar/src/distributed.rs
kfernandez31: True.

seastar/src/distributed.rs
kfernandez31: I believe this comment:
```
/// One can only delegate a "mutating map" on an instance if all other "mutating maps" on it have ceased.
/// In other words, you must ensure sequential acquisition of mutable borrows to an instance.
/// Failing to comply with this rule will lead to a `panic!`, much in the same way that breaking the rule for
/// a `RefCell` would do.
```
kind of explains it, but I'll add another one over the locks just in case.

kfernandez31:
v3: further comment resolution
@piodul review ping

piodul (approved):

